PR #19, opened 2021-01-23
Add did key and did web examples

OR13:
***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 23, 2021, 10:44 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fdid-test-suite%2F56fd379907847a1a69e6d5c2629759909254f4a7%2Findex.html%3FisPreview%3Dtrue)

```

Specification: https://rawcdn.githack.com/w3c/did-test-suite/56fd379907847a1a69e6d5c2629759909254f4a7/index.html?isPreview=true&publishDate=2021-01-23
ReSpec version: 26.0.0
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:217:19)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:106:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:208:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:87:18)

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/did-test-suite%2319.)._
</details>

### DIFF
--- a/packages/did-core-test-server/generate-latest-respec-data.js
+++ b/packages/did-core-test-server/generate-latest-respec-data.js
@@ -1,10 +1,52 @@
 const fs = require('fs');
+const path = require('path');
+const cheerio = require('cheerio');
+const moment = require('moment');
 const getReportResults = require('./services/getReportResults');
 const suitesInput = require('./suites/did-spec/default.json');
 
+const respecPath = path.resolve(__dirname, '../../index.html');
+
+const updateRespec = (suitesInput, suitesReportJson) => {
+  const spec = fs.readFileSync(respecPath).toString();
+  const $ = cheerio.load(spec);
+
+  $('#did-spec-suite-input').replaceWith(
+    `<pre
+        id="did-spec-suite-input"
+        class="example"
+        title="did-spec latest input report"
+      >
+${JSON.stringify(suitesInput, null, 2)}
+      </pre>`
+  );
+
+  $('#did-spec-suite-output').replaceWith(
+    `<pre
+        id="did-spec-suite-output"
+        class="example"
+        title="did-spec latest output report"
+      >
+${JSON.stringify(suitesReportJson, null, 2)}
+      </pre>`
+  );
+
+  $('#did-spec-report-date').replaceWith(
+    `<p id="did-spec-report-date" class="note">
+These test results were last generated 
+<span>
+${moment().format('LLLL')}
+</span>
+      </p>`
+  );
+
+  const updatedSpec = $.html();
+  fs.writeFileSync(respecPath, updatedSpec);
+};
+
 (async () => {
   const { suitesReportJson } = await getReportResults(suitesInput);
-
+  updateRespec(suitesInput, suitesReportJson);
   fs.writeFileSync(
     `../../test-vectors/did-spec.latest.json`,
     JSON.stringify(suitesReportJson, null, 2)
--- a/packages/did-core-test-server/suites/did-spec/default.json
+++ b/packages/did-core-test-server/suites/did-spec/default.json
@@ -2,53 +2,59 @@
   {
     "name": "did-spec",
     "didMethods": {
-      "did:example": {
+      "did:key": {
         "supportedContentTypes": [
           "application/did+json",
           "application/did+ld+json"
         ],
-        "dids": ["did:example:123"],
-        "didParameters": {
-          "hl": "did:example:123?hl=zQmWvQxTqbG2Z9HPJgG57jjwR154cKhbtJenbyYTWkjgF3e",
-          "service": "did:example:123?service=agent&relativeRef=%2Fpath%2Fto%2Fresource",
-          "relative-ref": "did:example:123?service=agent&relativeRef=%2Fpath%2Fto%2Fresource",
-          "version-id": "did:example:123?version-id=0.1.0",
-          "version-time": "did:example:123?version-time=1601151242"
-        },
-        "did:example:123": {
+        "dids": ["did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"],
+        "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB": {
           "application/did+json": {
             "didDocument": {
-              "@context": ["https://www.w3.org/ns/did/v1"],
-              "id": "did:example:123",
-              "canonicalId": "did:example:one-two-three",
-              "equivalentId": "did:example:one-two-three",
+              "@context": [
+                "https://www.w3.org/ns/did/v1",
+                {
+                  "@base": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+                }
+              ],
+              "id": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
               "verificationMethod": [
                 {
-                  "id": "#key-0",
+                  "id": "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
                   "type": "JsonWebKey2020",
-                  "controller": "did:example:123",
+                  "controller": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
                   "publicKeyJwk": {
-                    "kty": "OKP",
                     "crv": "Ed25519",
-                    "x": "VDXDwuGKVq91zxU6q7__jLDUq8_C5cuxECgd-1feFTE"
+                    "x": "LKacPea7et-QMhcyS02ZjsTHEVadsCj_XNnfFJYbwHo",
+                    "kty": "OKP"
                   }
                 },
                 {
-                  "id": "#key-1",
+                  "id": "#z6LSh8yKFPF2bLnD9mHQZMJL6zQQZsowYUeUaDd39rMLHBBx",
                   "type": "JsonWebKey2020",
-                  "controller": "did:example:123",
+                  "controller": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
                   "publicKeyJwk": {
                     "kty": "OKP",
                     "crv": "X25519",
-                    "x": "3kY9jl1by7pLzgJktUH-e9H6fihdVUb00-sTzkfmIl8"
+                    "x": "USe317eZOgxr3-neAHaJ8fZ056kb7miDxZvb5ymFKVM"
                   }
                 }
               ],
-              "authentication": ["#key-0"],
-              "assertionMethod": ["#key-0"],
-              "capabilityInvocation": ["#key-0"],
-              "capabilityDelegation": ["#key-0"],
-              "keyAgreement": ["#key-1"]
+              "authentication": [
+                "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+              ],
+              "assertionMethod": [
+                "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+              ],
+              "capabilityInvocation": [
+                "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+              ],
+              "capabilityDelegation": [
+                "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+              ],
+              "keyAgreement": [
+                "#z6LSh8yKFPF2bLnD9mHQZMJL6zQQZsowYUeUaDd39rMLHBBx"
+              ]
             },
             "didDocumentMetaData": {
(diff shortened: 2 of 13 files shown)


peacekeeper (changes requested):
This should be changed so that the examples for `application/did+json` don't contain `@context`, for reasons that have been discussed many times. It makes no sense to inject a JSON-LD specific entry into a plain JSON DID document.

OR13:
@peacekeeper take a look at https://did.key.transmute.industries/

https://github.com/transmute-industries/did-key.js

Are you saying that you intend to block implementations that are legal, but don't look the way that you want?

Why? 

Isn't the point of tests to show what is legal?

I am happy to include tests that don't contain an `@context` if you can find my an implementation of did-key (afaik there are none) or did-web (trivial for you to create) that does....

You are also welcome to create examples that look the way you want, but please don't block a PR that shows that an existing implementation is conformant, because you don't like the way the implementation shows conformance.

This will quickly escalate and slow down an already slow process of writing tests.... if you want to block legality of implementations `did-core` is the place to do that, not here.

These tests are not mean to be "whatever the editors want"... they are meant to reflect actual implementations and their support.

As an implementer of many did methods, I don't agree with your perspective, and I have written lots of code which intentionally does not support your perspective, and that code is legal. You are are welcome to write code that embodies your perspective and add examples of it here.

msporny:
> This should be changed so that the examples for application/did+json don't contain @context, for reasons that have been discussed many times. It makes no sense to inject a JSON-LD specific entry into a plain JSON DID document.

Digital Bazaar's DID Document producer implementations will contain `@context` for JSON-only representations. We do this because we use the same JSON Schema file to check for the correctness of `application/did+json` and `application/did+ld+json`. 

Our experience is that developers often get content negotiation wrong and we want to provide a single serialization that will work for JSON (just ignore but preserve `@context`) and JSON-LD (which requires that `@context` is defined). Having two code paths that only differ in whether or not @context is generated also creates the additional issue of creating more code to audit (which increases what can go wrong in complex systems).

packages/did-core-test-server/generate-latest-respec-data.js
```diff
@@ -1,10 +1,52 @@
 const fs = require('fs');
+const path = require('path');
+const cheerio = require('cheerio');
+const moment = require('moment');
 const getReportResults = require('./services/getReportResults');
 const suitesInput = require('./suites/did-spec/default.json');
 
+const respecPath = path.resolve(__dirname, '../../index.html');
+
+const updateRespec = (suitesInput, suitesReportJson) => {
```
msporny, line 10: This is some wizardry right here -- I like it -- wonder if any other W3C WG is doing stuff like this, feels like a useful design pattern.

packages/did-core-test-server/suites/did-spec/default.json
```diff
@@ -2,53 +2,59 @@
...
             "didDocument": {
-              "@context": ["https://www.w3.org/ns/did/v1"],
-              "id": "did:example:123",
-              "canonicalId": "did:example:one-two-three",
-              "equivalentId": "did:example:one-two-three",
+              "@context": [
+                "https://www.w3.org/ns/did/v1",
+                {
+                  "@base": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+                }
```
msporny, line 18: This makes me sad -- feels like something the did-key library implementation should do by setting the base instead of it being hard coded in the `@context`.

packages/did-core-test-server/suites/did-spec/default.json
```diff
@@ -2,53 +2,59 @@
...
+              ],
+              "id": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
               "verificationMethod": [
                 {
-                  "id": "#key-0",
+                  "id": "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
```
msporny, line 23: Hrm, why relative identifiers in did:key? This is more of a did:key quibble -- I had thought that everyone was going to use fully expanded identifiers (because there is no reason to optimize for space in did:key, right?)

packages/did-core-test-server/suites/did-spec/default.json
```diff
@@ -2,53 +2,59 @@
...
                   }
                 }
               ],
-              "authentication": ["#key-0"],
-              "assertionMethod": ["#key-0"],
-              "capabilityInvocation": ["#key-0"],
-              "capabilityDelegation": ["#key-0"],
-              "keyAgreement": ["#key-1"]
+              "authentication": [
+                "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
```
msporny, line 44: Again, why not fully expanded?

packages/did-core-test-server/suites/did-spec/default.json
```diff
@@ -2,53 +2,59 @@
...
                     "crv": "Ed25519",
-                    "x": "VDXDwuGKVq91zxU6q7__jLDUq8_C5cuxECgd-1feFTE"
+                    "x": "LKacPea7et-QMhcyS02ZjsTHEVadsCj_XNnfFJYbwHo",
+                    "kty": "OKP"
                   }
                 },
                 {
-                  "id": "#key-1",
+                  "id": "#z6LSh8yKFPF2bLnD9mHQZMJL6zQQZsowYUeUaDd39rMLHBBx",
                   "type": "JsonWebKey2020",
```
msporny, line 34: Assume that the did-key library will take an input parameter to specify the cryptosuite to use to express the key material?

packages/did-core-test-server/suites/methods/did-key/defaultSuiteConfig.json
```diff
@@ -0,0 +1,110 @@
...
+              "https://www.w3.org/ns/did/v1",
+              {
+                "@base": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB"
+              }
+            ],
+            "id": "did:key:z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
+            "verificationMethod": [
+              {
+                "id": "#z6MkhTZLPqg7kYwBe6y8utZZr1NfMViqXWep7rY7ogUSPgsB",
+                "type": "Ed25519VerificationKey2018",
```
msporny, line 75: Just as a heads-up, we're probably going to submit a `Ed25519VerificationKey2020` for did:key with `publicKeyMultibase` -- I assume the test suite allows for multiple did:key submissions?

packages/did-core-test-server/suites/methods/did-key/did-key.spec.js
```diff
@@ -0,0 +1,27 @@
+let { suiteConfig } = global;
```
msporny, line 1: This is about the name of the file... probably want did-key-transmute.spec.js -- as DB is going to want to submit their own as well.

msporny (approved):
Ok to merge... I have concerns about some of the stuff in did:key and did:web -- BUT, those are more concerns around the DID Method and not concerns about the test suite. Thanks for the addition of your implementations, @OR13! :)

packages/did-core-test-server/suites/did-spec/default.json
dlongley: Yeah, -1 to relative identifiers -- at least not as the main example as it encourages people to choose the format that I expect will frustrate devs the most. If we really need this for some reason, it should be its own special case example.

packages/did-core-test-server/suites/methods/did-key/did-key.spec.js
OR13: @msporny yes, this part of writing tests remains undefined.... IMO, the method conformance tests should ideally be a single collection, and represent all legal examples produced by real world implementations, for did-key, the problem is that implementations don't support all representations.... this can be solved by splitting the did-key by implementation, which appears to be in line with what @peacekeeper  wants.

OR13:
In summary, I think the main issue which remains unaddressed is how multiple implementations can be submitted for the same method.

The current structure allows for this to easily be tested, but in terms of a single test suite in which all methods are tested there is a problem.

The current tests assume that it is the DID Method, not the implementation that supports representations.... meaning in order for an implementation to show conformance it must either support all representations supported by other implementations, or be tested in isolation.... Its obviously better if we can use his to make methods support representations consistently... but I have low confidence that implementers have the time or skill needed to accomplish this.... so perhaps we should refactor the did core test suite to be implementation oriented as opposed to method oriented.

packages/did-core-test-server/suites/did-spec/default.json
OR13: our implementation is here: https://github.com/transmute-industries/did-key.js

essentially you create a keypair instance, which stores an internal representation of the key as buffers and meta data, and then you can ask for a `did+json` or `did+ld+json` representation of any given did-key.

the identifier remains the the same for both, according to the did key method spec, but they are represented differently depending on the caller's preference for JWK vs base58

packages/did-core-test-server/suites/did-spec/default.json
OR13: Our implementation prefers relative IRIs always, they are equivalent when compared via n-quads.

OR13: The real reason is that we want a did-key representation that looks like sidetree based dids, for testing... similar to the use of did-key with did:v1 I imagine.

As long as our implementation and yours produce the same n-quads we consider the implementations to be equivalent.

We also write tests using VCs and resolvers to show that they implementations both work, even though the look different.

packages/did-core-test-server/suites/did-spec/default.json
OR13: Our implementation prefers relative IRIs, and we ALWAYs try and optimize for space. We have tests the prove our relative ref implementation if consistent and equivalent with the expanded representation. 

Another reason we prefer the relative form is to match the structure of DIDs returned by sidetree based methods.

OR13: regarding the "main example" for a did method... which implementation is "the main one" remains an area for the test suite to address... my preference is that anything that is legal for the did method be tested, and therefore that implementations share a spec input for the method.... this will yield better alignment.

packages/did-core-test-server/suites/did-spec/default.json
OR13: It is extraneous information most of the time... but experience with JSON-LD has taught us that it is better to be explicit about this sort of thing.

packages/did-core-test-server/generate-latest-respec-data.js
OR13: agreed :)

OR13:
* so perhaps we should refactor the did core test suite to be implementation oriented as opposed to method oriented. * any objections?

csuwildcat:
No objections here. Thank you for writing tests.

OR13:
I plan on refactoring this to be "vendor oriented"... instead of "method oriented"... this will mean that for any given method, like `did:key` each vendor will be able to show a different level of conformance to did core... so some vendors will support `did+cbor` whereas others won't etc...

TallTed:
I suggest that the refactoring not be to "vendor oriented" but to "*implementation* oriented" as was previously expressed as the revised plan, replacing "[DID] method oriented".

It's entirely possible (though perhaps unlikely during our CR period[s]) that one vendor will have multiple implementations, probably covering different DID methods, but possibly even multiple implementations on one DID method.

This suggestion may be what you meant in recent comments, but I thought it worth certainty, instead of waiting until a lot of work had gone into refactoring to "vendor orientation" and then needing that work to be redone for "implementation orientation".

peacekeeper:
> Are you saying that you intend to block implementations that are legal, but don't look the way that you want?

> The intention of the test suite is to test what actual implementations are doing

The test suite should be built around what the spec says, not around what certain opinionated implementations are doing.

Yes your implementations that inject `@context` into application/did+json is "legal", but it's not the most direct and intutive way of implementing the data model and the JSON representation.

Our production rules for JSON-LD say "inject `@context`". Our production rules for JSON don't say that. So why do you do it there?

The reason why you're doing this is because you don't like the ADM and you would rather just have JSON-LD. I would also rather just have JSON-LD! But now that we have the ADM and multiple representations, we should use them in the test suite as intended. Your argument "it's legal" is not wrong, but it undermines the intentions of the specification, it ignores many F2F meetings and special topic calls that led to the current data model, and will confuse everyone who looks at the tests. You're trying to force your disagreement with the ADM design into the test suite using - no offence - a defiant and shallow argument.

Maybe at a later point in time we can add those special cases where additional entries like `@context` or `foo` or whatever are injected into application/did+json, but initially the test suite should be based on the spec in a straightforward way.

So -1 to merging this as-is.

PR #20, opened 2021-01-30
Refactor test suite to be vendor oriented and add example

OR13:
This PR refactors the test suite to be vendor oriented and adds support for transmute (my company) as an example vendor.

we currently support a number of did methods, including did:key and did:web.

This PR also adds examples of did documents produces by our code.

I have also removed the hypothetical test cases for `did:example`... since that is not a real did method.

Other vendors may choose to implement support for real did methods differently.

Our did key implementation is specifically designed to be easily integrated with did:web and sidetree based did methods, you can see some examples here:

- https://github.com/w3c-ccg/did.actor/blob/master/mike/did.json
- https://github.com/decentralized-identity/sidetree/blob/master/tests/vectors/resolution/afterUpdate.json
- https://github.com/transmute-industries/did-key.js/tree/master/packages/test-vectors/src/did-core-conformance
- https://staging.element.transmute.industries/workbench/resolver/did:elem:ropsten:EiCnhBQAmgedjtaxY3TYng-xlnH7Q-cfOsDR2v1gkXpwvA
- `https://did.key.transmute.industries/did:key:z6MkrkBrRLEDFTrf6MHT4i8FvN74LjYs86wKMt7L3FvYfN9Y`

since @peacekeeper objected to https://github.com/w3c/did-test-suite/pull/19

I would like to point out that @selfissued @csuwildcat @troyronda @tplooker @kdenhartog and other sidetree wg members also rely on the did document structural examples I have provided, i helped write the specs and implementations these methods are built on.

If you don't like our implementations or opinions on JSON / JSON-LD support perhaps you should consider implementing a did method that embodies your preferences.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 11, 2021, 11:06 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fdid-test-suite%2F0c9bb8b04b295dafeb127c561c7d05c996634b46%2Findex.html%3FisPreview%3Dtrue)

```

Specification: https://rawcdn.githack.com/w3c/did-test-suite/0c9bb8b04b295dafeb127c561c7d05c996634b46/index.html?isPreview=true&publishDate=2021-02-11
ReSpec version: 26.0.0
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:217:19)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:106:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:208:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:87:18)

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/did-test-suite%2320.)._
</details>


csuwildcat:
Looks good to me!

Therecanbeonlyone1969:
Looks good to me! I support the merge of this PR!

msporny (approved):
Giant PR, too big to review line-by-line, but understand that this is a major refactor, so probably can't be broken down easily.

+1 to merge as this implements a direction that the WG needs in order to demonstrate what vendors are actually doing in the field.

troyronda:
LGTM. Thanks @OR13.

peacekeeper (changes requested):
There's a lot of good work in here, thanks @OR13 !

But -1 to merging as-is, I'll write an explanation in the next few days.

msporny:
> But -1 to merging as-is, I'll write an explanation in the next few days.

Hmm, this is a strange objection. What @OR13 is implementing is a fairly standard way of doing implementation test suites at W3C. Implementations are expected to conform to most of the specification, but DO NOT have to conform to ALL of the specification. They are meant to be a sanity check wrt. whether the WG has imposed normative requirements that implementers are willing to implement (or not).

There is nothing (AFAICT) that Transmute has implemented in their implementations that go against the normative requirements in the specification, so blocking the PR based on what they've implemented isn't appropriate. Even if Transmute's implementation were to violate the specification in some way, it's important for the WG to know that... especially if other implementers choose to violate the specification in the same way. At the end of the day, specifications need to match reality lest they become a work of fiction (which tend to be useless to most everyone in the real world).

TallTed:
(@OR13 - I'm going with your description here; I've not reviewed the code in your PR to confirm whether it is actually now "vendor oriented".)

As I've said elsewhere, "vendor oriented" is just as much the wrong tactic as "method oriented".

*Implementation* oriented is the correct tactic, because any given vendor *may* implement multiple methods, which implementation *may* be as multiple methods in one implementation (a/k/a software application/package) *or* as multiple implementations with one method per implementation.

A given implementation should be testable for all (individuated) features of any or all methods.

This gives us checkmarks for each feature, and in aggregate for each method.  These checkmarks are what I expect we will decide to use as the exit criteria for CR, and also for removing (or confirming as viable) the sections/features now marked as AT RISK.  Results might also give some scale or degree measure, other than the simple yes/no of checkmarks.

This also gives deployers/users/customers a clear basic way to compare and contrast implementations, as they may not care about support for some optional feature(s), which they can combine with the DID Rubric and other factors in choosing what to deploy in their environment.

OR13:
@TallTed I think your language is in line with this PR and its changes.

This PR adds support for:

1. multiple vendors
2. multiple implementations supported by each vendor

This PR removes the implication that all vendors implement a method the same way, which is obviously false :)

haardikk21:
LGTM!

In respect to the debate about JSON-LD vs JSON, injecting `@context` values into JSON is much more favorable - for uses with Sidetree, and for other implementors using DID's and/or VC's. The specification does not block putting values like `@context` into a JSON only document, and it is easier and far more usable to treat these documents as JSON.

IMO, JSON-LD is an extension of JSON, and if it can be represented and used as JSON in a easier, more usable way, while also conforming to the spec, there is no reason to treat it differently.

mprorock:
@peacekeeper can you elaborate on what is holding this PR up? we are looking to test against and utilize the implementation provided by @OR13, but need it in the mainline test suites to validate consistently across multiple vendors and various implementations.

TallTed:
I suggest that this PR's title be changed to more accurately reflect what it primarily does (i.e., "refactors the test suite to be implementation-oriented" since it can't be vendor-oriented by virtue of handling "multiple implementations supported by each vendor"), instead of being specifically Transmute-focused.

There should be no need to "add support for Transmute" nor any other vendor, nor to incorporate Transmute-(nor any other vendor)-specific content which may have the unintended consequence of making Transmute (or such other vendor) appear as special and/or recommended in some way(s) vs any other vendors whose test results are or are not included in our eventual Implementation Report.

Therefore, I also suggest removing any Transmute-specific content from this PR except insofar as that content is submission of initial test suite results and/or sanity-testing of the test suite -- which may, and likely should, include Transmute having "failure" marks on some tests, as this suite _**MUST**_ be entirely vendor-agnostic.

(It would be best to submit any Transmute test results as a separate PR for inclusion in the Implementation Report, not directly to the test suite itself.)

OR13:
@TallTed

> Therefore, I also suggest removing any Transmute-specific content from this PR except insofar as that content is submission of initial test suite results and/or sanity-testing of the test suite -- which may, and likely should, include Transmute having "failure" marks on some tests, as this suite _**MUST**_ be entirely vendor-agnostic.

The suite IS vendor agnostic, thats what this refactor accomplishes... I needed to add least 1 vendor to show that it worked... I am not inclined to spend any more time on tests until this is merged... we need to start somewhere, and continuing to add fake data that passes the tests is not as helpful as providing real world data extracted from real world implementations.

msporny:
> The suite IS vendor agnostic, thats what this refactor accomplishes... I needed to add at least 1 vendor to show that it worked.

Yes, +1 -- Digital Bazaar is now hard blocked on this PR being merged. We want to contribute and can't until this is merged. 

Can we merge this -- it's a dramatic improvement and necessary for us to get the test suite into a shape where we can parallelize test suite work.

OR13:
@peacekeeper 

See 

Note that Transmute has chosen to preserve `@context` in our `did+json` implementation of did-key, for the sake of testing sidetree style did documents, without needing to make network requests... we are not doing this because we "love linked data" (we do)... we are doing it to make testing easier for us, and because we think its a good thing... you are welcome to implement did-key `did+json` that is not linked data, and if folks prefer your implementation to ours, they are free to use yours.

TallTed:
(I don't know what objection(s) @peacekeeper has.  I'm writing about my own concerns here.)

@OR13 
> I needed to add at least 1 vendor to show that it worked...

I strongly urge you to stop referring to "vendor" when "implementation" is what is meant.

Each _implementation_ should result in a column of passes-and-fails in an Implementation Report.

Any given _vendor_ might have zero, one, or multiple implementations reflected in such an Implementation Report.

When building a test suite, it is surely helpful to have an implementation handy which can serve or provide data for a "test suite test case", and that implementation will surely come from some source, commonly referred to as a vendor.

But neither that implementation nor its vendor should be immortalized anywhere in the test suite itself.

The branch you're merging from here is fortunately ephemeral, but its name is problematic nonetheless: `feat/add-transmute-vendor`.

As far as I can tell, contrary to what you said recently, this refactor is _not_ implementation focused; it's strongly _vendor_ focused, and with Transmute being the only vendor named in the test suite source, Transmute _appears_ to be special.

`packages/did-core-test-server/suites/vendors/README.md` says "This folder is for testing vendors in isolation."

I can't think of any scenario under which a vendor with multiple implementations would want to test all of them at once.  (In parallel or in sequence, sure, but not as a block, which is what appears to me to be meant.)  I expect any such vendor would want to test each *implementation* in isolation.

Likewise, I can't think of any scenario where anyone would want to test all implementations from all vendors at once.  (Again, in parallel or in sequence, sure.)  I can think of many where people would want to *vier* all test results from all implementations from all vendors ... and that is typically the Implementation Report, which *may* be set up to accept future test results as well as those which are produced during CR/PR phases leading up to TR.

You have said that it is not your intent to make this test suite vendor-specific nor vendor-focused, but your intent is not carrying through, to my eye.

"Transmute" occurs 175+ times in [the changes](https://github.com/w3c/did-test-suite/pull/20/files) I can quickly see.  (I do all my github work through their web interface, so I can't do a lot of the analysis that might otherwise be desirable.)  A fair number of those are within the `index.html` file which I *think* is a results report.  A bunch more are just repetition of the transmute folder in the `.../suites/vendors/...` filename list.

But then there are an awful lot of replacements of `did-spec` with `transmute`, and I cannot tell how much of that simply reflects "system under test" (and so should not be seen by anyone else running the suite on their implementation) and how much does not.

csuwildcat:
Clearly @OR13 is engaged in a stealthy attempt to insert Transmute everywhere in this test system and exemplar implementation test in order to subliminally influence everyone who views it with these Transmute references, reeling them into Transmutes maniacal world DIDomination plan. I don't know about you, but I'm sure glad we caught this in time before any unsuspecting people got #594aa8-purple pilled.

TallTed:
It is important to remember that participants in the DID WG and CCG are not the only people who will be consuming what we produce, and what we think of as blatantly obvious is not always so.

peacekeeper:
This PR like several others before does NOT accurately reflect the intention of the data model and its different representations.

This is most obvious in places [like this](https://github.com/w3c/did-test-suite/pull/20/files#diff-a8a3554961ed29e5b1350d7b79d523a982939c10cab2d515a6f28c689dc15a83), where the `application/did+json` and `application/did+ld+json` representations are simply set to the same file.

@OR13 previously had a PR that wanted to define the JSON representation as the byte-to-byte equivalent to the JSON-LD representation. The code above tries to promote this intent. However, the consensus in the WG was that the JSON and JSON-LD representations are NOT meant to be byte-to-byte equivalent.

I think the code in **did-json-production.js** and **did-ld-json-production.js** doesn't actually test production of those representations.

What the code should test is the following sequence of steps:

1. Create a DID document abstract data model (e.g. by calling the `resolve()` function, or in some other way). This abstract data model should only contain properties such as `verificationMethod` or `service` from the [Core Properties](https://w3c.github.io/did-core/#core-properties) section, but it should not contain any representation-specific entries.
2. Execute the production rules from each one of the conformant representations and then test the resulting representation.

In this case, there will clearly be NO `@context` in the `application/did+json` representation, since the production rules of that representation don't mention it. This is also what the diagram in https://github.com/w3c/did-core/pull/596 illustrates. This is not a matter of perspective, and it is not something everyone can implement however they want.

The Transmute implementation is doing something else. I think what it does instead is EITHER:

1. Create a DID document abstract data model (e.g. by calling the `resolve()` function, or in some other way).
2. Execute the production rules for `application/did+ld+json`.
3. Execute the consumption rules for `application/did+ld+json`.
4. Execute the production rules for `application/did+json`. In this case `@context` will be present in the result since it was "preserved" in the previous step. This sequence is what we have been calling "conversion" between representations.

OR:

1. Call the `resolveRepresentation()` function with `«[ "accept" → "application/did+json" ]»`, and for some reason the DID method implementation injects `@context` as an extension into the DID document even though it's not required in the requested representation. I would agree that it is "legal" for Transmute to do that if they choose to (even though I find it not very intuitive).

In both cases, the `application/did+json` representation will have an `@context`. I would agree that in both cases, no normative statements were violated. But in both cases, steps were executed that are different and more complex than straightforward production of a representation from a simple DID document.

I'm not blocking this because I don't like Transmute's implementation-specific choices. I'm blocking this because the test suite doesn't actually test what it says it's testing.

-1 until we're all aligned on this.

TallTed:
@peacekeeper -- If I'm reading your objection correctly, it's to a small number of tests, not to the suite as a whole? I suggest that you adjust your objection, to insert a flagging comment in those few tests, which links to your objection here, which can then lead to changes of those few tests while the *other* tests can be put to preliminary use (including having other tests submitted). Alternatively, you could accede to merging this PR, and create issues about those particular tests immediately thereafter.

@OR13, all -- I won't block merging of this PR, but my concerns noted above have not been resolved nor eased, and these concerns will need to be addressed in the very near future.  Again:

1. Vendor-orientation is not the right approach for this nor any specification's test suite. Implementation-orientation is.
2. Sample data should not reflect any vendor's branding. (Submitting results of running the test suite against an implementation as "initial results for the implementation report" and/or "testing the test suite against an implementation" would be fine, but that does not seem to me to be what's being done here.)
3. @peacekeeper has flagged some tests which appear not to have been written to match the spec, but to match the particular implementation which has been used to test the test suite (which implementation appears not to conform to the spec, as it stands). This underscores my concerns.

OR13:
@peacekeeper our implementation of did-key and our implementation of sidetree both have this same behavior and are legal according to did core, if you don't like our implementation, don't use it....  

> I would agree that in both cases, no normative statements were violated.

This ^ is the only thing that matters for the test suite.... your opinions on our software implementations are welcome on the repos that hold the implementations... feel free to raise an issue here:

- https://github.com/transmute-industries/did-key.js
- https://github.com/transmute-industries/sidetree.js
- https://github.com/decentralized-identity/sidetree

@brentzundel @burnburn can we please arrange a special topic call we have a special topic call to unblock the test suite.

OR13:
> 1. Vendor-orientation is not the right approach for this nor any specification's test suite. Implementation-orientation is.

This was discussed on a a call, and I find the distinction between a vendor and an implementation not helpful... for example, there are many vendors that all rely on hyperledger ursa.... are we going to say that only 1 implementation of of hyperledger ursa crypto is to be tested? which one? rust, python, go?

> 2\. Sample data should not reflect any vendor's branding. (Submitting results of running the test suite against an implementation as "initial results for the implementation report" and/or "testing the test suite against an implementation" would be fine, but that does not seem to me to be what's being done here.)

I am not sure what you mean by "branding"... on the previous WG call I was directed to refactor the tests to allow multiple vendors to submit sample implementation data, I have done that.


> 3\. @peacekeeper has flagged some tests which appear not to have been written to match the spec, but to match the particular implementation which has been used to test the test suite (which implementation appears not to conform to the spec, as it stands). This underscores my concerns.

Yes, those tests are for boolean and numbers... which the spec has 0 examples of... making them really hard to test.... However, with the recent changes to deactivated I believe we will be able to test `deactivated` boolean flag in resolution response... there will still be 0 examples of numbers and booleans in did document... I don't know how to fix that issue, but it will remain an issue... this is the kind of feedback that we need on these tests...

TallTed:
> > 1. Vendor-orientation is not the right approach for this nor any specification's test suite. Implementation-orientation is.
> 
> This was discussed on a a call, and I find the distinction between a vendor and an implementation not helpful... for example, there are many vendors that all rely on hyperledger ursa.... are we going to say that only 1 implementation of of hyperledger ursa crypto is to be tested? which one? rust, python, go?

This paragraph suggests a deeper misunderstanding than I thought was in play.

No, we are not "going to say that only 1 implementation of of hyperledger ursa crypto is to be tested".  We are going to (be able to) test (and record results for) *each implementation* -- rust, python, go, java, c, etc.  But again, we're not going to distinguish by *language* because there may be multiple _implementations_ produced using any given language. (And we're not testing "hyperledger ursa crypto" per se, because that's not what we're specifying; we're specifying DID Core.)

A "vendor" is a person or company which produces and makes available some software package.

An "implementation" is some collection of code which makes available some feature(s) within a software package.

One "vendor" may actually have hundreds of "implementations" over time.

My employer is a vendor.  My employer produces multiple products which do not always share code.  Each product might include a different implementation (by which I mean different code which may have been written by different programmers and/or in different programming languages) of DID Core.  Each implementation should be testable, and testing each implementation (each product, in this case) should result in a different column in the test results report.  There should not be one column for my employer, as a vendor, that blends results from all products/implementations; there should be multiple columns, one for each product/implementation.

> > 2. Sample data should not reflect any vendor's branding. (Submitting results of running the test suite against an implementation as "initial results for the implementation report" and/or "testing the test suite against an implementation" would be fine, but that does not seem to me to be what's being done here.)
> 
> I am not sure what you mean by "branding"... on the previous WG call I was directed to refactor the tests to allow multiple vendors to submit sample implementation data, I have done that.

What I mean by "branding" is having a vendor's or product's name (a/k/a brand) embedded in the test suite or specification.  I would be voicing just as much concern if any other vendor or product were named throughout, including that of, or one from, my employer.

Yes, multiple vendors must be able to submit implementation test results. Each vendor should also be able to submit multiple implementation test results for multiple products, with each product's results being distinct from the other(s).  A given vendor might want to (and should be able to) submit test results for multiple versions of the same product, and results for each version should be distinguishable from the other version(s); for example, VendorX might submit test results for Product1v1.0 which show failure for tests 123–127, and later submit rest results for Product1v1.1 which show success for tests 123–125 and failure for 126–127, and still later submit results for Product1v1.2 which show success for tests 123–127.

(Note that conformance testing is not performed *only* during the specification process! Such tests are *often* run long after the specification is final, to demonstrate superiority of one product/implementation over another.  These might be from competing vendors, or at different price/feature points from one vendor.)

peacekeeper:
> If I'm reading your objection correctly, it's to a small number of tests, not to the suite as a whole?

@TallTed yes that's correct, no objection to the test suite as a whole.

>  I suggest that you adjust your objection, to insert a flagging comment

The problem with this is that once the controversial tests are merged, implementers may not read or understand the objection and start implementing by just looking at the examples. It may then be very difficult to fix the tests and implementations later.

Instead, I suggest removing the affected tests (**did-resolution**, **did-json-production**, **did-ld-json-production**) plus the vendor examples and other content related to those tests.

This way we could merge the refactored test suite itself (which is great work!) as well as non-controversial tests such as **did-syntax** and **did-parameters**, and discuss the others separately.

OR13:
@peacekeeper In order to test resolution we need JSON objects that map include data we expect to have in resolution.... in other words, you are actually objecting to the way we are testing resolution... (significantly more helpful than objecting to the way a vendor implemented a did method).

Since I've been using this JSON structure for a while now to test both sidetree based methods and did key, I thought it would be useful starting place... and hopefully make it easier for Microsoft, SecureKey and any other folks who are familiar with sidetree to add test for their methods... 

You should feel free to implement resolution tests however you like in another PR, and we can review it... or we can merge this PR and you can refactor the tests (not our implementation fixtures) to test resolution however you want and I will then update our fixtures to pass your new test structure.

What I have included in this PR works for us, but there are other ways that it could work... I am not opposed to implementing things differently... I am opposed to blocking a PR because you don't like the way the test were written, but spending most of the comments saying you don't like the vendors implementation of their did method... it sorta comes across as not constructive for the test suite....

We really need to pickup the pace on these tests, and you have now held up this PR for 12 days... 

If we merge it, I would be more than happy to review a refactor done by you that changes everything, and if I need to generate new JSON to pass your new test structure I am fine with that...

We need to make progress on the actual tests, not bicker over who's method includes an `@context`...

I propose we do one of the following:

1. merge the PR and wait for you to do a refactoring 
2. close the PR and wait for you to add resolution tests for a did method you are the implementer of

TallTed:
(Sorry, I started writing this an hour before @OR13's [comment](https://github.com/w3c/did-test-suite/pull/20#issuecomment-776773125) was posted, and got pulled into another task...)

@peacekeeper 
> Instead, I suggest removing the affected tests (**did-resolution**, **did-json-production**, **did-ld-json-production**) plus the vendor examples and other content related to those tests.

This is along the same lines as my suggestion to insert flags on the specific tests to which you object, as I would expect and recommend that the content which is removed from this PR would immediately form the basis for one or more new PRs on which the discussion can be more focused -- and this PR #20 could then be merged and allow other work to proceed.

However, _"the vendor examples and other content related to those tests"_ is vague enough that I (and, I suspect, many if not all others) could not easily remove it from this PR.  The more easy such changes can be made for the PR author, the more likely they will happen.

Perhaps you can make a "review" pass over this PR which includes deletion of (1) the tests you identified above and (2) the vendor examples and other content related to those tests.  Those can then be revised and re-added through new PRs (probably best to have one PR for each of the three tests named above, though one PR including all 3 might be workable; and then a separate PR for the vendor example(s) [from Transmute and/or other vendor(s)] on that and any other tests). I'd suggest that @OR13 is best to create the new PR with the Transmute submission.

OR13:
@TallTed thanks for your proposal. @peacekeeper if I delete the resolution tests will you accept this PR?

peacekeeper:
@OR13 @TallTed Yes, as I already suggested above, please update the PR to exclude the affected tests (did-resolution, did-json-production, did-ld-json-production) plus the vendor examples and other content related to those tests.

After that I'm fine with merging (but let me review first after you update the PR). And then you can create another PR to add those tests back in (which I will probably object to again for the reasons stated above, but at least we'll have some progress).

> you have now held up this PR for 12 days

Please don't try to pressure me like that. I reviewed the PR and came to the conclusion that a part of the test suite doesn't actually test what it says it's testing, which would be harmful if merged. How is that my fault?

> spending most of the comments saying you don't like the vendors implementation of their did method

Please read my comments above. I'm not blocking it because of this.

OR13:
@peacekeeper I removed the effected tests in https://github.com/w3c/did-test-suite/pull/20/commits/19d8d7492eafe5f6190a293a40d200ffa8051ef0

peacekeeper:
Please also remove the examples related to production and resolution, they are misleading and can cause confusion.

Also, there's still code that sets `application/did+json` to be equal to `application/did+ld+json`.

Issue #22, opened 2021-02-22
Need tests for DID Parameters


msporny:
[3.2.1 DID Parameters](https://w3c.github.io/did-core/#did-parameters)

<table>
  <tr>
   <td><strong>Parameter Name</strong>
   </td>
   <td><strong>Description</strong>
   </td>
  </tr>
  <tr>
   <td><code><a href="https://sites.local/did-core/#dfn-service">service</a></code>
   </td>
   <td>
If present, the associated value <em>MUST</em> be an <a href="https://infra.spec.whatwg.org/#ascii-string">ASCII string</a>.
   </td>
  </tr>
  <tr>
   <td><code>relativeRef</code>
   </td>
   <td>
If present, the associated value <em>MUST</em> be an <a href="https://infra.spec.whatwg.org/#ascii-string">ASCII string</a> and <em>MUST</em> use percent-encoding for certain characters as specified in <a href="https://tools.ietf.org/html/rfc3986#section-2.1">RFC3986 Section 2.1</a>.
   </td>
  </tr>
  <tr>
   <td><code>versionId</code>
   </td>
   <td>
If present, the associated value <em>MUST</em> be an <a href="https://infra.spec.whatwg.org/#ascii-string">ASCII string</a>.
   </td>
  </tr>
  <tr>
   <td><code>versionTime</code>
   </td>
   <td>
If present, the associated value <em>MUST</em> be an <a href="https://infra.spec.whatwg.org/#ascii-string">ASCII string</a> which is a valid XML datetime value, as defined in section 3.3.7 of <a href="https://www.w3.org/TR/xmlschema11-2/">W3C XML Schema Definition Language (XSD) 1.1 Part 2: Datatypes</a> [<a href="https://sites.local/did-core/#bib-xmlschema11-2">XMLSCHEMA11-2</a>]. 
<p>

This datetime value <em>MUST</em> be normalized to UTC 00:00:00 and without sub-second decimal precision. For example: <code>2020-12-20T19:17:47Z</code>.
   </td>
  </tr>
  <tr>
   <td><code>hl</code>
   </td>
   <td>
If present, the associated value <em>MUST</em> be an <a href="https://infra.spec.whatwg.org/#ascii-string">ASCII string</a>.
   </td>
  </tr>
</table>

wyc:
@clehner will be implementing this on behalf of @spruceid.

msporny:
[...]

PR #34, opened 2021-03-08
Test DID Parameters

clehner:
This is aimed at fixing #22.
- Update DID parameters to use camelCase, following https://github.com/w3c/did-core/pull/553.
- Test `versionTime` DID parameter. Update test case to use [xsd:dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime) rather than UNIX timestamp, to match DID Core.
- Check that DID parameter values are ASCII only as specified in DID Core, rather than for all possible DID parameters.

Notes/questions:
- The `versionTime` test validates the parameter with a regex based on the [lexical representation of xsd:dateTime](https://www.w3.org/TR/xmlschema11-2/#nt-dateTimeRep) with the constraints for `versionTime` applied. But the [day-of-month constraints](https://www.w3.org/TR/xmlschema11-2/#con-dateTime-dayValue) are not checked (e.g. when is February 29th a valid date). Should we check these day-of-month constraints too?
- Should more exhaustive tests be done for percent-encoding in `relativeRef`?
- Should negative test cases be added? We could add an area in the input config [didParameters](https://github.com/spruceid/did-test-suite/blob/9025962c284115321afa5125612e474b815d4f18/packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json#L10) object for "bad" DID URLs, and check that they do not pass validation.
- Should all the DID parameter tests be applied to each tested DID URL? Currently the tests for `relativeRef` and `versionTime` are only applied to their respectively labeled test cases in the [config](https://github.com/spruceid/did-test-suite/blob/9025962c284115321afa5125612e474b815d4f18/packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json#L10-L16). This means there could be invalid uses of these parameters in the other DID URLs which would not be detected here. I wonder if this could be a source for confusion. We could instead apply all DID parameter tests to all DID URLs.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 19, 2021, 5:27 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fspruceid%2Fdid-test-suite%2F7925112b59a2f92ee820d0afe03034e9f05af3c9%2Findex.html%3FisPreview%3Dtrue)

```

Specification: https://rawcdn.githack.com/spruceid/did-test-suite/7925112b59a2f92ee820d0afe03034e9f05af3c9/index.html?isPreview=true&publishDate=2021-03-19
ReSpec version: 26.3.0
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:218:19)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:107:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:207:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:86:18)

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/did-test-suite%2334.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -401,7 +401,7 @@ it('The DID method name MUST be an ASCII lowercase string.', () => {
               "did-spec",
               "did:example",
               "did-parameters",
-              "relative-ref"
+              "relativeRef"
             ],
             "title": "The associated value MUST be an ASCII string.",
             "status": "passed"
@@ -411,7 +411,7 @@ it('The DID method name MUST be an ASCII lowercase string.', () => {
               "did-spec",
               "did:example",
               "did-parameters",
-              "relative-ref"
+              "relativeRef"
             ],
             "title": "MUST use percent-encoding for certain characters as specified in RFC3986 Section 2.1.",
             "status": "passed"
@@ -421,7 +421,7 @@ it('The DID method name MUST be an ASCII lowercase string.', () => {
               "did-spec",
               "did:example",
               "did-parameters",
-              "version-id"
+              "versionId"
             ],
             "title": "The associated value MUST be an ASCII string.",
             "status": "passed"
@@ -431,7 +431,7 @@ it('The DID method name MUST be an ASCII lowercase string.', () => {
               "did-spec",
               "did:example",
               "did-parameters",
-              "version-time"
+              "versionTime"
             ],
             "title": "The associated value MUST be an ASCII string.",
             "status": "passed"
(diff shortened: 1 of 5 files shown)


OR13 (approved):

msporny (approved):

OR13:
>  Should we check these day-of-month constraints too?

imo, no need.

> Should more exhaustive tests be done for percent-encoding in relativeRef?

yes

> Should negative test cases be added? 

yes

> Should all the DID parameter tests be applied to each tested DID URL? 

not sure, but I suspect the answer is no... ideally we would have clean cases for "good" and "bad" examples, and very focused tests.

peacekeeper (approved):

OR13:
We agreed to merge this ASAP.

PR #37, opened 2021-03-19
Feat/did document metadata

oed:
This PR mainly adds tests for the following **DID Document Metadata** properties:
* created
* updated
* nextUpdate
* versionId
* nextVersionId

In addition it fixed a few problems with the testsuite:
* `content-type` -> `contentType`
* `contentType` should be on `didResolutionMetadata`
* `canonicalId` and `equivalentId` should be in `didDocumentMetadata`
* `versionTime` param should be XML Datetime

PR #38, opened 2021-03-20
Add first implementation of did:key and Core Properties tests.

msporny:


packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json
```diff
@@ -1,6 +1,49 @@
 {
   "name": "did-spec",
   "didMethods": {
+    "did:key - 2018 Ed25519 cryptosuite - did-key-js - Digital Bazaar": {
```
msporny, line 4: @OR13 -- I expect we're going to have to label things like this given that there will be multiple implementations and types of did:key tests. Thoughts?

packages/did-core-test-server/suites/did-spec/did-core-properties.js
```diff
@@ -0,0 +1,31 @@
+const utils = require('./utils');
+const {isValidDID} = utils;
+
+const jsonMediaTypes = ['application/did+ld+json', 'application/did+json'];
+
+const generateDidCorePropertiesTests = ({did, resolutionResult}) => {
+  const {didDocument} = resolutionResult;
+
+  it('5.1.1 DID Subject - The value of id MUST be a string that ' +
```
msporny, line 9: @OR13 -- Thinking of prefacing tests with the section they're in to make it easier to map these sorts of things to sections of the specification. Thoughts?

packages/did-core-test-server/suites/did-spec/did-core-properties.js
```diff
@@ -0,0 +1,31 @@
...
+    'conforms to the rules in § 3.1 DID Syntax and MUST exist in the root ' +
+    'map of the data model for the DID document.', async () => {
+      expect(didDocument).toHaveProperty('id');
+      expect(isValidDID(didDocument.id)).toBe(true);
```
msporny, line 13: Jest's expect syntax leaves something to be desired... not a fan.

packages/did-core-test-server/suites/did-spec/did-core-properties.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+}
+
+const didCorePropertiesTests = (suiteConfig) => {
+  describe('did-core-properties', () => {
+    suiteConfig.dids.forEach((did) => {
+      describe(did, () => {
+        for(const [mediaType, resolutionResult] of Object.entries(suiteConfig[did])) {
+          if(jsonMediaTypes.includes(mediaType)) {
+            generateDidCorePropertiesTests({did, resolutionResult});
```
msporny, line 23: Deep nesting of tests seems to be unavoidable. :(

packages/did-core-test-server/suites/did-spec/utils.js
```diff
@@ -18,7 +18,14 @@ const isValidURL = (data) => {
   return valid;
 };
 
+const isValidDID = (did) => {
```
msporny, line 21: Disagree with uppercase "DID", and "URL" in function names. Keeping it like this to go with the established code practices in this repo.

packages/did-core-test-server/suites/did-spec/utils.js
```diff
@@ -18,7 +18,14 @@ const isValidURL = (data) => {
...
+  const didRegex = /did:(?<method>[a-z0-9]+):(?<idchar>[a-zA-Z0-9:\\-_]+)/;
```
msporny, line 22: Feels like there is going to be a ton of duplication w/ this test suite and the Transmute one: https://github.com/transmute-industries/did-core/blob/main/packages/did-core-tests/json-schemas/did.json

Are there any plans to merge these two things?

packages/did-core-test-server/suites/did-spec/default.json
```diff
@@ -2,6 +2,49 @@
   {
     "name": "did-spec",
     "didMethods": {
+      "did:key - 2018 Ed25519 cryptosuite - did-key-js - Digital Bazaar": {
```
msporny, line 5: I expect this file to get very unwieldy -- we have multiple implementations coming down the pipeline -- can we break it up into implementations?

test-vectors/did-spec.latest.json
```diff
@@ -2,6 +2,68 @@
...
+          "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH"
+        ],
+        "title": "The DID method name MUST NOT be empty.",
+        "status": "passed"
+      },
+      {
+        "ancestors": [
+          "did-spec",
+          "did:key - 2018 Ed25519 cryptosuite - did-key-js - Digital Bazaar",
+          "5. Core Properties",
```
msporny, line 49: @OR13 is this an issue? I'd like us to group tests by section of the spec, if possible. Make it obvious which tests are testing which part of the spec. Speak clearly to which DID Methods pass/fail which specific statements and make it easy for developers to find these items.

msporny:
Oh no, Merge is supported for this repo -- please don't do regular merges. We need to disable this -- @iherman could you please either 1) give the CODEOWNERs access to the settings for this repo... OR, disable regular merges, and make sure to only allow Squash and Rebase merges.

iherman:
> Oh no, Merge is supported for this repo -- please don't do regular merges. We need to disable this -- @iherman could you please either 1) give the CODEOWNERs access to the settings for this repo... OR, 2) disable regular merges, and make sure to only allow Squash and Rebase merges.

I have disabled merge commits. The setup seems to be identical to the did-core settings now...

packages/did-core-test-server/suites/did-spec/did-core-properties.js
OR13: yeah that sounds good. I wrote this respec to jest code gen here: https://github.com/transmute-industries/did-core/tree/main/packages/did-core-tests/__generated__

which automatically handled all this stuff, but it harder to do that for tests related to did methods vs the spec itself.

packages/did-core-test-server/suites/did-spec/did-core-properties.js
OR13: I'm a fan! https://jestjs.io/docs/expect see also https://github.com/jest-community/jest-extended

The line you hilighted is asserting the return value of a helper function... you could rewrite the helper function to `assertValidDID` and have it throw instead of return a boolean.

packages/did-core-test-server/suites/did-spec/did-core-properties.js
OR13: yep, and problem gets worse the more we attempt to have all did methods share the same configs... lots of conditional nested code paths.

packages/did-core-test-server/suites/did-spec/utils.js
OR13: yeah, i think the official recomendation is to use `Did` or `Url` when you are defining a function / method... obviously the ship has sailed for `new URL("https://example.com")`

packages/did-core-test-server/suites/did-spec/utils.js
OR13: I can published the package and we can use it as a dependency here... but I'm not going to try to maintain our test suite in a W3C repo governed by W3C Consensus :)

packages/did-core-test-server/suites/did-spec/default.json
OR13: yeah, we may want to switch to `.js` and use require to build the json instead... like we have done here:

https://github.com/w3c-ccg/vc-http-api/blob/main/packages/vc-http-api-test-server/config/Transmute.js#L43

test-vectors/did-spec.latest.json
OR13: I don't think its an issue, but like most software I expect the test suite to change as people try to actually use it.

OR13 (approved):

packages/did-core-test-server/suites/did-spec/did-core-properties.js
msporny: > you could rewrite the helper function to assertValidDID and have it throw instead of return a boolean.

Yes, as @shigeya raised in an issue, that should probably be what we do to make the tests easier to read. Granted, I'm not looking forward to going back and rewriting lots of tests. :)

packages/did-core-test-server/suites/did-spec/utils.js
msporny: > yeah, i think the official recomendation is to use Did or Url

Would you object to a global search/replace to move to that model? I'm concerned that people might look at the test suite and then duplicate the mistake in their code.

I can do the search/replace if you don't object.

packages/did-core-test-server/suites/did-spec/utils.js
msporny: > I'm not going to try to maintain our test suite in a W3C repo governed by W3C Consensus

What could possibly go wrong? :P 

Don't worry about it -- I expect we'll be able to get through it w/o the JSON Schema stuff... but would have probably made the tests easier/nicer than just dumping stuff in `util.js`.

packages/did-core-test-server/suites/did-spec/default.json
msporny: > yeah, we may want to switch to `.js` and use require to build the json instead...

+1, let's do that before we get too many new entries and it becomes major surgery.

packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json
OR13: yeah, probably there is better ways of handling this, but for now, this is good.

packages/did-core-test-server/suites/did-spec/utils.js
shigeya: I'm not too fond of duplication 😝    I think having a common library in this package is a good thing. Anyone who doesn't want to depends on this package can create a similar function in their package.

packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json
```diff
@@ -1,6 +1,49 @@
...
+            ],
+            "keyAgreement": [{
+              "id": "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH#zBzoR5sqFgi6q3iFia8JPNfENCpi7RNSTKF7XNXX96SBY4",
+              "type": "X25519KeyAgreementKey2019",
+              "controller": "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH",
+              "publicKeyBase58": "JhNWeSVLMYccCk7iopQW4guaSJTojqpMEELgSLhKwRr"
+            }]
+          },
+          "didDocumentMetaData": {
+            "content-type": "application/did+ld+json"
```
peacekeeper, line 41: This is DID Resolution Metadata, not DID Document Metadata.

peacekeeper: Also, it should be `contentType`, not `content-type`.

peacekeeper:
Just a small comment, I think it would be nicer if the did:key configuration is added after did:example, not inserted before it.

Also, wouldn't it be better to have implementation-specific test inputs/outputs in separate files, or do we want to have everything in a single gigantic "defaultSuiteConfig.json" of "default.json" file (and what's the difference between these two anyway?)

packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json
msporny: This fix is being dealt with in PR #37.

msporny:
> Just a small comment, I think it would be nicer if the did:key configuration is added after did:example, not inserted before it.

Ok, I can move that.

> Also, wouldn't it be better to have implementation-specific test inputs/outputs in separate files, or do we want to have everything in a single gigantic "defaultSuiteConfig.json" of "default.json" file (and what's the difference between these two anyway?)

Agree that we want implementation-specific tests in separate files, that was discussed in https://github.com/w3c/did-test-suite/pull/38#discussion_r598914926.

@OR13, I'm also confused about the difference of `defaultSuiteConfig.json` and `default.json`... I know one of them is used to generate the report (which I have no idea how I get my new tests to show up in that file, help @OR13), and the other is used when doing the command-line tests. I'd prefer that we just had one default config.

msporny:
> > Just a small comment, I think it would be nicer if the did:key configuration is added after did:example, not inserted before it.
> 
> Ok, I can move that.

Done in 424cce527d42ec7de01deb5ae2bc2362f83467f2.

I believe this PR is ready to merge. I'll merge it in 24 hours unless there are objections.

packages/did-core-test-server/suites/did-spec/did-core-properties.js
```diff
@@ -0,0 +1,382 @@
...
+
+  return verificationMethods;
+};
+
+const didCorePropertiesTests = (suiteConfig) => {
+  describe('did-core-properties', () => {
+    suiteConfig.dids.forEach((did) => {
+      describe(did, () => {
+        for(const [mediaType, resolutionResult] of Object.entries(suiteConfig[did])) {
+          if(jsonMediaTypes.includes(mediaType)) {
```
peacekeeper, line 373: What does this do? It feels strange that testing of Core Properties somehow depends on media types?

packages/did-core-test-server/suites/did-spec/did-core-properties.js
```diff
@@ -0,0 +1,382 @@
+const utils = require('./utils');
+const {
+  isValidDID, isValidURI, isValidBase58, isValidJwk, isValidVerificationMethod,
+  getAbsoluteDIDURL
+} = utils;
+const jose = require('node-jose');
+
+const jsonMediaTypes = ['application/did+ld+json', 'application/did+json'];
```
peacekeeper, line 8: What does this do? See comment below

shigeya:
Is it reasonable to merge method-specific configuration?

Do we want to use a similar pattern as we separate DID Core and DID registry?

If everybody who register DID method touching configuration in this repository too, we need to accept PR to both this code and DID registry, that's not a great idea.

It may be possible to have a JSON object in DID registry, which this code (optionally) read. Then, we can reduce the number of merges.

packages/did-core-test-server/suites/did-spec/did-core-properties.js
OR13: We discussed this issue on the working group call, its an artifact of the fact that we are:

1. using json fixtures for the ADM
2. testing serialization of the ADM in JSON (did+ld+json and did+json)

I spent a good amount of time thinking about how to make this better without increasing the burden on test implementers, and this is what I came up with.

packages/did-core-test-server/suites/did-spec/did-core-properties.js
OR13: https://github.com/w3c/did-test-suite/pull/38#discussion_r601585709

OR13:
> Is it reasonable to merge method-specific configuration?
> 

Yes, but we may refactor in the future, once more contributors have figured out how this works.

> Do we want to use a similar pattern as we separate DID Core and DID registry?
> 

Yes, imo we need to handle code splitting and json object merging from smaller files to make this easier to use... PRs welcome.

> If everybody who register DID method touching configuration in this repository too, we need to accept PR to both this code and DID registry, that's not a great idea.
> 

The did spec registries only registers did methods, so its not really a good example.... https://github.com/w3c-ccg/vc-http-api is a better example since it has apis for "issuer", "holder", "verifier"... and you may not care to test all 3, and you may not care to use features of any given one.... replace "issuer", "holder", "verifier" with 'did+json', 'did+ld+json' and 'did+cbor', and internal features such as 'revocation list' with 'canonicalId' or 'service'.

> It may be possible to have a JSON object in DID registry, which this code (optionally) read. Then, we can reduce the number of merges.

-1 to splitting test fixtures across repos until they are stable, and even then, probably not a good idea.

shigeya:
> Yes, but we may refactor in the future, once more contributors have figured out how this works.

> Yes, imo we need to handle code splitting and json object merging from smaller files to make this easier to use... PRs welcome.

> -1 to splitting test fixtures across repos until they are stable, and even then, probably not a good idea.

As far as we plan to refactor later, I fully agree. We want to minimize dependency.

> The did spec registries only registers did methods, so its not really a good example

I understand. Thanks.

Issue #24, opened 2021-02-22
Need test for Conforming Producer


msporny:
6.1 Production and Consumption - A conforming producer MUST take a DID document data model and a representation-specific entries map as input into the production process. The conforming producer MAY accept additional options as input into the production process.

6.1 Production and Consumption - A conforming producer MUST serialize all entries in the DID document data model, and the representation-specific entries map, that do not have explicit processing rules for the representation being produced using only the representation's data type processing rules and return the serialization after the production process completes.

6.1 Production and Consumption - A conforming producer MUST return the Media Type string associated with the representation after the production process completes.

6.1 Production and Consumption -  A conforming producer MUST NOT produce non-conforming DIDs or DID documents.

msporny:
PR #41 has been raised to address this issue. This issue should be closed once that PR has been merged.

peacekeeper:
The normative statements listed in this issue need to be updated to match what's in the CR spec now (especially section 6.1):

https://www.w3.org/TR/2021/CR-did-core-20210318/#production-and-consumption

msporny:
> The normative statements listed in this issue need to be updated to match what's in the CR spec now (especially section 6.1):
> 
> https://www.w3.org/TR/2021/CR-did-core-20210318/#production-and-consumption

I have just updated the normative statements for this issue to match what's in the spec.

msporny:
This has been done, but needs refinement. The refinement is being tracked in #48, continuing there.

Issue #32, opened 2021-02-25
We need to refactor fixtures for implementers


OR13:
We need to clearly show the difference between abstract and concrete representations.

there should be did document data models for both abstract and concrete examples

OR13:
I am asserting that no PRs will make it in until we have at least one example of an implementer fixture that shows how to do this... and its going to be dangerous to skip over it.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-03-30-did#section7-3) on 2021-03-30

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 7.3. We need to refactor fixtures for implementers 

_See github issue [did-test-suite#32](https://github.com/w3c/did-test-suite/issues/32)._




**Orie Steele:** Same point Markus and others have raised. Essentially, the structure of the test suites is not set in stone or finalized or working.  
… This issue is raised at a time of great frustration, but since then we have got many more people contributing to this.  
… I don't think we have finished refactoring the test fixtures, as evidenced by Markus's comments, but I think we are on our way.  

**Brent Zundel:** This one is unassigned  

**Manu Sporny:** I will take it  
… Critical path... I don't want to put myself in the critical path... I will make an attempt.  
… There is an order of operations here. One is we figure out details of testing all this stuff. Once we figure it out, it will be a bit of a mess  
… We will need to refactor it. Then we expect the DID method implementers to come in and fill it out.  
… Concerned about order of operations...  
… If anyone gets to it before I do, please take a shot at it.  

> *Markus Sabadello:* Please review my PR about DID Resolution tests.. [https://github.com/w3c/did-test-suite/pull/43](https://github.com/w3c/did-test-suite/pull/43)

</details>

msporny:
I have raised PR #52 to address this issue.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-04-06-did#section7-7) on 2021-04-06

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 7.7. need clear instructions for implementers 

_See github issue [did-test-suite#33](https://github.com/w3c/did-test-suite/issues/33), [did-test-suite#32](https://github.com/w3c/did-test-suite/issues/32)._




**Markus Sabadello:** also related to refactoring  
… implementers of a did method or the spec in another way so they know how exactly they submit their data  

[...]

PR #41, opened 2021-03-21
Add Producer tests for Section 1.4 and Section 6.1.

msporny:
Implement the tests needed by issue #24.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 24, 2021, 2:01 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fdid-test-suite%2F37057d1553566e0b38f2bf21e49bb3b8c0ce8373%2Findex.html%3FisPreview%3Dtrue)

```

Specification: https://rawcdn.githack.com/w3c/did-test-suite/37057d1553566e0b38f2bf21e49bb3b8c0ce8373/index.html?isPreview=true&publishDate=2021-03-24
ReSpec version: 26.4.0
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:218:19)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:107:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:207:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:86:18)

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/did-test-suite%2341.)._
</details>

### DIFF
--- a/packages/did-core-test-server/suites/did-spec/did-producer.js
+++ b/packages/did-core-test-server/suites/did-spec/did-producer.js
@@ -6,8 +6,30 @@ const generateDidProducerTests = ({did, resolutionResult}) => {
   const {didDocument} = resolutionResult;
   const contentType = resolutionResult.didDocumentMetaData['content-type'];
 
-  it('1.4 Conformance - A conforming producer MUST NOT produce ' +
-    'non-conforming DIDs or DID documents.', async () => {
+  it('6.1 Production and Consumption - A conforming producer MUST take a ' +
+    'DID document data model and a representation-specific entries map as ' +
+    'input into the production process. The conforming producer MAY accept ' +
+    'additional options as input into the production process.', async () => {
+      if(jsonMediaTypes.includes('application/did+ld+json')) {
+        const context = didDocument['@context'];
+        if(typeof context === 'string') {
+          expect(context).toBe('https://www.w3.org/ns/did/v1');
+        } else if(Array.isArray(context)) {
+          expect(context[0]).toBe('https://www.w3.org/ns/did/v1');
+          reserializedContext = JSON.parse(JSON.stringify(context));
+          expect(deepEqual(context, reserializedContext)).toBe(true);
+        } else {
+          throw new Error('Invalid @context value '+ context);
+        }
+      }
+  });
+
+  it('6.1 Production and Consumption - A conforming producer MUST ' +
+    'serialize all entries in the DID document data model, and the ' +
+    'representation-specific entries map, that do not have explicit ' +
+    'processing rules for the representation being produced using only the ' +
+    'representation\'s data type processing rules and return the ' +
+    'serialization after the production process completes.', async () => {
       if(jsonMediaTypes.includes(contentType)) {
         reserializedDidDocument = JSON.parse(JSON.stringify(didDocument));
         expect(deepEqual(didDocument, reserializedDidDocument)).toBe(true);
@@ -16,22 +38,20 @@ const generateDidProducerTests = ({did, resolutionResult}) => {
       }
   });
 
-  it('6.1 Production and Consumption - A conforming producer MUST serialize ' +
-    'all entries in the data model that do not have explicit processing ' +
-    'rules for the representation being produced using only the ' +
-    'representation\'s data type processing rules.', async () => {
-      if(jsonMediaTypes.includes(contentType)) {
-        reserializedDidDocument = JSON.parse(JSON.stringify(didDocument));
-        expect(deepEqual(didDocument, reserializedDidDocument)).toBe(true);
-      } else {
-        throw new Error('Unknown producer for content-type: '+ contentType);
-      }
-  });
-
-  it('6.1 Production and Consumption - A conforming producer MUST indicate ' +
-    'which representation has been used for a DID document via a Media Type ' +
-    'as described in § 7.1.2 DID Resolution Metadata.', async () => {
+  it('6.1 Production and Consumption - A conforming producer MUST return ' +
+    'the Media Type string associated with the representation after the ' +
+    'production process completes.', async () => {
       expect(typeof contentType === 'string').toBe(true);
+  });
+
+  it('6.1 Production and Consumption - A conforming producer MUST NOT ' +
+    'produce non-conforming DIDs or DID documents.', async () => {
+      if(jsonMediaTypes.includes(contentType)) {
+        reserializedDidDocument = JSON.parse(JSON.stringify(didDocument));
+        expect(deepEqual(didDocument, reserializedDidDocument)).toBe(true);
+      } else {
+        throw new Error('Unknown producer for content-type: '+ contentType);
+      }
   });
 }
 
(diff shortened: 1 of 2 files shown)


OR13 (approved):

peacekeeper (changes requested):
I think this is a good start, but the tested normative statements don't seem to match what's in the CR spec now, especially section 6.1.

E.g. this PR tests the following normative statement:

> A conforming producer MUST serialize all entries in the data model that do not have explicit processing rules for the representation being produced using only the representation's data type processing rules.

But the CR spec contains the following normative statement:

> A conforming producer MUST serialize all entries in the DID document data model, and the representation-specific entries map, that do not have explicit processing rules for the representation being produced using only the representation's data type processing rules and return the serialization after the production process completes.

msporny:
> the tested normative statements don't seem to match what's in the CR spec now

Doh! Thank you @peacekeeper, good catch. I forgot that that entire section got a rewrite and we needed to go back and update the normative statements in the issue. Working on updating the PR now.

msporny:
> Working on updating the issue and the PR now.

This has been done now. Please re-review @peacekeeper .

packages/did-core-test-server/suites/did-spec/did-producer.js
```diff
@@ -0,0 +1,70 @@
+const utils = require('./utils');
+const jsonMediaTypes = ['application/did+ld+json', 'application/did+json'];
+const deepEqual = require('deep-equal')
+
+const generateDidProducerTests = ({did, resolutionResult}) => {
+  const {didDocument} = resolutionResult;
+  const contentType = resolutionResult.didDocumentMetaData['content-type'];
```
peacekeeper, line 7:
```suggestion
  const contentType = resolutionResult.didResolutionMetaData['contentType'];
```

packages/did-core-test-server/suites/did-spec/did-producer.js
```diff
@@ -0,0 +1,70 @@
...
+
+  it('6.1 Production and Consumption - A conforming producer MUST take a ' +
+    'DID document data model and a representation-specific entries map as ' +
+    'input into the production process. The conforming producer MAY accept ' +
+    'additional options as input into the production process.', async () => {
+      if(jsonMediaTypes.includes('application/did+ld+json')) {
```
peacekeeper, line 13:
```suggestion
      if(jsonMediaTypes.includes(contentType)) {
```

Is this what you meant?

packages/did-core-test-server/suites/did-spec/did-producer.js
```diff
@@ -0,0 +1,70 @@
...
+        expect(deepEqual(didDocument, reserializedDidDocument)).toBe(true);
+      } else {
+        throw new Error('Unknown producer for content-type: '+ contentType);
+      }
+  });
+
+  it('6.1 Production and Consumption - A conforming producer MUST return ' +
+    'the Media Type string associated with the representation after the ' +
+    'production process completes.', async () => {
+      expect(typeof contentType === 'string').toBe(true);
```
peacekeeper, line 44: This feels insufficient.. How does this test the statement above?

peacekeeper (commented):
I still don't think this is testing production properly. The input data should distinguish better between the abstract data model (expressed as JSON objects), the representation-specific map (expressed as JSON objects), and the representations (serialized JSON, JSON-LD, hex-encoded CBOR), and then the production/consumption tests should be written based on that.

E.g. see how this data structure for DID Resolution distinguishes between abstract data model and representations: https://github.com/w3c/did-test-suite/blob/4f9b23506933bc77906281f6a4e05c2283c69917/packages/did-core-test-server/suites/did-resolution/defaultSuiteConfig.json

packages/did-core-test-server/suites/did-spec/did-producer.js
OR13: should be `contentType === application/did+json`, or similar.

packages/did-core-test-server/suites/did-spec/did-producer.js
msporny: Good catch, no, I meant `contentType === 'application/did+ld+json'` ... fixed in c0fb4a72b515e5157bd7b2e3dca93d2f41aa7292.

The reason the test only tests 'application/did+ld+json' is because that's the only time where you know that there is a representation-specific entries map.

packages/did-core-test-server/suites/did-spec/did-producer.js
msporny: The normative statement is just saying that a contentType has to exist and the spec defines contentType as being a string. That's what the test is testing.

> should be contentType === application/did+json, or similar.

Remember that this test runs /for all producers/. So, if the test was the above... any non-JSON representation would fail the test, which is not what we want. We just want to make sure producers are returning a media type string for a representation.

peacekeeper: The normative statement says much more here than just "the content type MUST be a string", which is what the code is testing....

msporny: > The normative statement says much more here than just "the content type MUST be a string", which is what the code is testing....

What test would you write for this normative statement?

msporny:
> I think the approach for testing production should be that we have "abstract data model things" in the input data, and "representation things" in the output data, and the tests should run based on that.

Ok, I'll take a look at those PRs and see what I can do.

packages/did-core-test-server/suites/did-spec/did-producer.js
```diff
@@ -0,0 +1,70 @@
...
+      }
+  });
+}
+
+const didProducerTests = (suiteConfig) => {
+  describe('did-producer', () => {
+    suiteConfig.dids.forEach((did) => {
+      describe(did, () => {
+        for(const [mediaType, resolutionResult] of Object.entries(suiteConfig[did])) {
+          generateDidProducerTests({did, resolutionResult});
```
OR13, line 63: on the call, we discussed how we think this may be better to be a function of the ADM instead of the resolution result.

peacekeeper:
For example look at what @clehner did here: https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json#L108

And I did something similar here: https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/did-resolution/defaultSuiteConfig.json#L63, https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/did-resolution/defaultSuiteConfig.json#L84

There's a clear distinction between abstract data model and representations.

packages/did-core-test-server/suites/did-spec/did-producer.js
OR13: I think we may wish to address these concerns in an issue before attempting changes to the tests proposed in the PR

https://github.com/w3c/did-test-suite/issues/48

peacekeeper (approved):
@msporny we discussed this on a call today, and since @OR13 just created an issue, I'd be fine with merging this PR now and then improving later..

peacekeeper:
Also related: https://github.com/w3c/did-test-suite/issues/32

PR #42, opened 2021-03-24
Fix some bugs in the example DID method.

peacekeeper:


packages/did-core-test-server/suites/did-spec/defaultSuiteConfig.json
```diff
@@ -94,10 +95,11 @@
...
             },
             "didDocumentMetaData": {
-              "content-type": "application/did+cool+json"
             },
-            "didResolutionMetaData": {}
-          }
+            "didResolutionMetaData": {
+              "contentType": "application/did+cool+json"
+            }
+        }
```
davidlehn, line 102:
```suggestion
          }
```

OR13:
There are now merge conflicts.

peacekeeper:
Merge conflicts resolved

msporny:
> Merge conflicts resolved

Theeeyyyy're baaaack! :)

I have rebased my commits too many times at this point... the test suite doesn't have good test separation, leading to folks stomping all over each other. I'm sure it'll be obvious how to structure the suite once we're done. :)

OR13:
yeah, this is good evidence that separate files would be better... we should build `defaultSuiteConfig.js` from multiple json.

peacekeeper:
Merge conflicts resolved again..

OR13 (changes requested):
@peacekeeper please confirm you understand that this change breaks compatibility with JSON-LD and verifiable credentials, and I will be happy to merge.

peacekeeper:
> this change breaks compatibility

The only change left in the PR (after resolving merge conflicts) is in the plain JSON example, how does that break anything in JSON-LD?

OR13:
@peacekeeper the way that JSON-LD verifiable credentials work assumes the presence of this now removed property... when we remove the verification method types form the `"@context": ["https://www.w3.org/ns/did/v1"],` this will become even more obvious, since `Ed25519VerificationKey2018` and `JsonWebKey2020` will no longer be "in the context without extension"... As I have noted over and over again, defaulting to a JSON representation without context is a direct attack on JSON-LD verifiable credential tooling....

https://en.wikipedia.org/wiki/Consequentialism

> The future amplification of the effects of small decisions[51] is an important factor that makes it more difficult to predict the ethical value of consequences,[52] even though most would agree that only predictable consequences are charged with a moral responsibility.[43]

I don't see the value in the change-set, and I can see it causing problems in the future.... in short your example is incompatible with VCs... for no good reason, since `did:example` isn't even a real did method.... if your goal was to submit a real did method that was intentionally incompatible with VCs I would be happy to approve this PR, but not in the abstract.

I want to hear you say: "I want the json representation example to be incompatible with JSON-LD"... If I approved this PR as is, it would be with the assumption that your intention is exactly this. Would you like me to approve?

peacekeeper:
@OR13 I really don't get how a DID document in the JSON representation can potentially break anything in VCs. If certain tooling only works with JSON-LD and requires `@context`, then that's fine, you don't have use the plain JSON representation (or XML or YAML, etc.) if your tooling doesn't work with those.

Would you say that the [JSON example](https://www.w3.org/TR/2021/CR-did-core-20210318/#example-21-example-did-document-in-json-representation) in the spec is also a "direct attack on JSON-LD verifiable credential tooling" ?

OR13 (approved):
For the record, I don't consider removing a property that increases interoperability fixing a bug.

Issue #31, opened 2021-02-22
Need tests for Metadata Structure


msporny:
[7.3 Metadata Structure \
](https://w3c.github.io/did-core/#metadata-structure)

The structure used to communicate this metadata _MUST_ be a [map](https://infra.spec.whatwg.org/#maps) of properties. 

Each property name _MUST_ be a [string](https://infra.spec.whatwg.org/#string). 

Each property value _MUST_ be a [string](https://infra.spec.whatwg.org/#string), [map](https://infra.spec.whatwg.org/#maps), [list](https://infra.spec.whatwg.org/#list), [ordered set](https://infra.spec.whatwg.org/#ordered-set), [boolean](https://infra.spec.whatwg.org/#boolean), or [null](https://infra.spec.whatwg.org/#null). 

The values within any complex data structures such as maps and lists _MUST_ be one of these data types as well. 

All metadata property definitions registered in the DID Specification Registries [[DID-SPEC-REGISTRIES]](https://infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes) _MUST_ define the value type, including any additional formats or restrictions to that value (for example, a string formatted as a date or as a decimal integer).

The entire metadata structure _MUST_ be serializable according to the [JSON serialization rules](https://infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes) in the [[INFRA]](https://www.w3.org/TR/2021/CR-did-core-20210318/#bib-infra) specification.

peacekeeper:
@msporny I edited your comment to reflect the latest version of the spec.

shigeya:
I'm wondering how we can test this statement:

> All metadata property definitions registered in the DID Specification Registries [DID-SPEC-REGISTRIES] MUST define the value type, including any additional formats or restrictions to that value (for example, a string formatted as a date or as a decimal integer).

How do we know property definitions on the test suite side? We just assume non-core properties to be in the spec registry?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-03-30-did#section7-4) on 2021-03-30

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 7.4. Need tests for Metadata Structure 

_See github issue [did-test-suite#31](https://github.com/w3c/did-test-suite/issues/31)._




**Shigeya Suzuki:** I almost finished writing tests besides one issue, related to the DID registry, as I commented in the issue.  
[...]

PR #47, opened 2021-03-30
DID Metadata Structure Test (first set)

shigeya:
This PR contains on-going work for Metadata Structure tests, Issue #31.

Includes most of the tests besides one with dependency with spec registry.

### DIFF
--- a/packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
+++ b/packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
@@ -1,3 +1,39 @@
+const didMetadataStructureTest_r = (didDocumentMetadata) => {
+    describe("didDocumentMetadata", () => {
+        it('The structure used to communicate this metadata MUST be a map of properties.', () => {
+            expect(didDocumentMetadata instanceof Object).toBeTruthy();
+        });
+
+        it('Each property name MUST be a string.', () => {
+            Object.keys(didDocumentMetadata).forEach((property_name) => {
+                expect(typeof property_name).toBe("string");
+            });
+        });
+
+        it('Each property value MUST be a string, map, list, ordered set, boolean, or null.', () => {
+            Object.keys(didDocumentMetadata).forEach((property_name) => {
+                expect(typeof didDocumentMetadata[property_name]).toEqual(expect.stringMatching(/(string|object|boolean)/));  // map, list, set and null are objects
+            });
+        });
+
+        it('The values within any complex data structures such as maps and lists ' +
+            'MUST be one of these data types as well.', () => {
+                Object.keys(didDocumentMetadata).forEach((property_name) => {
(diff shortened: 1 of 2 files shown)


packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
```diff
@@ -0,0 +1,61 @@
+const didMetadataStructureTest_r = (didDocumentMetadata) => {
+    describe("didDocumentMetadata", () => {
```
OR13, line 2: if you can prefix this with the section of the spec that will help us make sure we are covering everything

packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
```diff
@@ -0,0 +1,61 @@
...
+                expect(() => {
+                    obj = JSON.parse(JSON.stringify(didDocumentMetadata));
+                }).not.toThrow();
+                expect(didDocumentMetadata).toEqual(obj);
+        });
+    });
+};
+
+const didMetadataStructureTests = (suiteConfig) => {
+    describe('7.3 Metadata Structure', () => {
```
OR13, line 45: maybe this is sufficient

OR13 (approved):

packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
```diff
@@ -0,0 +1,61 @@
...
+
+        it('Each property name MUST be a string.', () => {
+            Object.keys(didDocumentMetadata).forEach((property_name) => {
+                expect(typeof property_name).toBe("string");
+            });
+        });
+
+        it('Each property value MUST be a string, map, list, ordered set, boolean, or null.', () => {
+            Object.keys(didDocumentMetadata).forEach((property_name) => {
+                expect(typeof didDocumentMetadata[property_name]).toEqual(expect.stringMatching(/(string|object|boolean)/));  // map, list, set and null are objects
```
OR13, line 15: a bit of a nit, but potentially we can test "list / set" better

clehner (approved):

peacekeeper (approved):

packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
shigeya: It's a function called from main `describe` section. For the source-code-wise, prefix might be useful, but if we ran test verbosely, the section number appears multiple times. May let me consider what is the best solution in future commits.

packages/did-core-test-server/suites/did-spec/did-metadata-structure.js
shigeya: Any suggestion welcome. I think we need to have some is-a functions for all of the basic types and consistently use it.

PR #52, opened 2021-04-03
Refactor the entire test suite to make it easier to contribute implementations.

msporny:
I have performed the following things to refactor the test suite:

* Split did-spec into suites associated with the major sections of the test suite: identifier, core properties, consumption, production, resolution, and dereferencing.
* Created an `implementations` directory that is shared among all the test suites. There can be 5 classes of implementations (so far) -- DID Methods, resolvers, dereferencers, and soon, producers and consumers.
* Renamed and restructured the test suite output to make it more logically grouped. The general order is: major section of the spec > implementation being tested > minor section of the spec > specific variation implementation being tested > specific normative statement. It's a bit awkward, but given that test suite implementers have chosen to test different sections of the spec in different ways, it's the only logical structuring that we can do, AFAICT.
* Unified the configuration files.

If we merge this in, it means that people submitting implementations only need to submit ONE `.json` file for each class of implementation they want to test. For example, if someone has a specific DID Method Document they want to submit to the test suite, all they have to do is submit *one* DID Document file in the implementations directory.

There are still things that need to be done:

* Test Suite Implementers (for example for production, consumption, resolution, or URL dereferencing) may want to further re-factor their test suites given the new structure. This is optional, and we'll only know after discussing the new layout with the implementers.
* I'm still not sure how we're generating reports from this and will have to chat with @OR13 about that.
* I'm almost certain the "test server" part of this is broken now, but shouldn't be hard to fix. I just don't know what the architecture and goals are there, again, will need to chat with @OR13 about that.


clehner (approved):

OR13:
We can scrub the test server and stick to CI for the report generation.

OR13 (approved):

PR #55, opened 2021-04-09
Separate data model from representations in tests.

msporny:
This PR separates the DID Document data model, representation-specific entries, and representations into distinct fields for input into, and output from the test suites. The production tests have been rewritten to take advantage of the new layout and more directly test the inputs/outputs.


packages/did-core-test-server/suites/implementations/did-example-didwg.json
```diff
@@ -83,6 +85,10 @@
         "capabilityDelegation": ["#key-0"],
         "keyAgreement": ["#key-1"]
       },
+      "dmRepresentationSpecificEntries": {
+        "@context": ["https://www.w3.org/ns/did/v1"]
+      },
+      "representation": "{\"@context\":[\"https://www.w3.org/ns/did/v1\"],\"id\":\"did:example:123\",\"verificationMethod\":[{\"id\":\"did:example:123#key-0\",\"type\":\"Ed25519VerificationKey2018\",\"controller\":\"did:example:123\",\"publicKeyBase58\":\"6fioC1zcDPyPEL19pXRS2E4iJ46zH7xP6uSgAaPdwDrx\"},{\"id\":\"did:example:123#key-1\",\"type\":\"X25519KeyAgreementKey2019\",\"controller\":\"did:example:123\",\"publicKeyBase58\":\"FxfdY3DCQxVZddKGAtSjZdFW9bCCW7oRwZn1NFJ2Tbg2\"}],\"authentication\":[\"#key-0\"],\"assertionMethod\":[\"#key-0\"],\"capabilityInvocation\":[\"#key-0\"],\"capabilityDelegation\":[\"#key-0\"],\"keyAgreement\":[\"#key-1\"]}",
```
clehner, line 91: Could `didDocumentStream` be used instead of `representation`, for consistency with DID Resolution (`resolveStream` function)?

clehner, line 91: If there are test inputs with more than one representation, would it make sense to move `dmProperties` and `dmRepresentationSpecificEntries` up a level (outside the representation's object)? Otherwise these would be duplicated if there are multiple representations for a DID. For example, the following would be to consider the data model as like a pseudo-representation:
```js
  "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH": {
    "application/did+ld+json": {
      ...
    }
    "didDocumentDataModel": {
      "dmProperties": {
        ...
      },
      "dmRepresentationSpecificEntries": {
        ...
      }
    }
  }
```

clehner (approved):

OR13 (approved):

OR13 (changes requested):
@msporny please apply the suggested changes from @clehner

packages/did-core-test-server/suites/implementations/did-example-didwg.json
peacekeeper: I think `representation` is more accurate, since `didDocumentStream` is only mentioned in the DID Resolution section, but this data here seems to be about production and consumption of representations.

msporny: Keep as is for now, raise an issue, and change it later if there is stronger consensus to change it to something else.

TODO: Create issue, link issue, merge PR.

msporny: Group decided to move `dmProperties` up a level (right under the did, at the same level as the media type).

peacekeeper (approved):
Fine with merging this after decisions in the issue are applied

packages/did-core-test-server/suites/implementations/did-example-didwg.json
OR13: @msporny I think actually its a mistake to move this up a layer, reading the spec:

https://w3c.github.io/did-core/#data-model

I would be in favor of this:

```
  "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH": {
    "application/did+json": {
      "didDocumentStream": "{\"id\":..." // note no context
      ...
    },
   "application/did+ld+json": {
       "didDocumentStream": "{\"@context\":..., \"id\":..." // note context
      ...
    }
    "didDocumentDataModel": {
      "dmProperties": {
        "id":...
        ...
      },
      "dmRepresentationSpecificEntries": {
         "@context":...
        ...
      }
    }
  }
```

@peacekeeper  thoughts?

peacekeeper: I think dmRepresentationSpecificEntries should be inside a representation's object, because (as the name says) it is a representation-specific map.

I.e. I think it should be like this:

```
  "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH": {
    "didDocumentDataModel": {
      "dmProperties": {
        ...
      },
    },
    "application/did+ld+json": {
      "didDocumentDataModel": {
        "dmRepresentationSpecificEntries": {
          "@context": ...
        }
      },
      "representation": "{\"@context\":..., \"id\":..."
    }
  }
```

msporny: Tracking issue created here: https://github.com/w3c/did-test-suite/issues/63

msporny: Applied in https://github.com/w3c/did-test-suite/pull/55/commits/56ac8522ad6997773321f64a3e4834b458ff91fb.

Tracking @OR13's alternate in issue https://github.com/w3c/did-test-suite/issues/64

PR #60, opened 2021-04-14
Fix to isXmlDatetime utility function to not allowing sub-seconds

shigeya:
The function definition includes sub-seconds but the spec doesn't allow sub-seconds.

Luckily, no tests using this predicate up to now.


shigeya:
Updated regexp since I forgot to remove one closing curly brace.

msporny (approved):

PR #66, opened 2021-04-18
Update identifier, core properties, and production suites to use new jest extensions

msporny:
I updated the identifier, core properties, and production suites to use @shigeya (AWESOME!!!) new jest extensions. I also tested to make sure they fail when given invalid input as I went along (and they did). I haven't removed the util.js files just yet... we need to move everything over to the new jest matchers before removing the old `util.js` test functions.

### DIFF
--- a/packages/did-core-test-server/suites/did-core-properties/did-core-properties.js
+++ b/packages/did-core-test-server/suites/did-core-properties/did-core-properties.js
@@ -15,7 +15,7 @@ const generateDidCorePropertiesTests = (
     'conforms to the rules in § 3.1 DID Syntax and MUST exist in the root ' +
     'map of the data model for the DID document.', async () => {
       expect(didDocument).toHaveProperty('id');
-      expect(isValidDID(didDocument.id)).toBe(true);
+      expect(didDocument.id).toBeValidDid();
   });
 
   it('5.1.2 DID Controller - The controller property is OPTIONAL. If ' +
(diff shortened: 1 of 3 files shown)


clehner (approved):

packages/did-core-test-server/suites/did-core-properties/did-core-properties.js
```diff
@@ -298,7 +298,7 @@ const generateDidCorePropertiesTests = (
           } else if(Array.isArray(serviceValue.type)) {
             const types = serviceValue.type;
             types.forEach(type => {
-              expect(typeof type === 'string').toBe(true);
```
shigeya, line 301: This can be replaced with `isInfraString` but looks like I failed to export the predicate. Will look into my code.

shigeya (commented):
Looks great. added a few suggestions.

I found my bug by local test for `is*` predicates. Will submit a PR to fix that later.

packages/did-core-test-server/suites/did-core-properties/did-core-properties.js
shigeya: No, not here. different `if`.

packages/did-core-test-server/suites/did-core-properties/did-core-properties.js
```diff
@@ -51,7 +51,7 @@ const generateDidCorePropertiesTests = (
       if(verificationMethod) {
         expect(Array.isArray(verificationMethod)).toBe(true);
```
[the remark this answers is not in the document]
shigeya: my bad. should be `toBeArray()`.

PR #71, opened 2021-04-27
Add initial tests for JSON production

csuwildcat:


OR13 (approved):

msporny:
@csuwildcat thanks for the start for these tests. The tests need to be done (completely and fully) by next Tuesday. If they are not, the JSON section will most likely be dropped from the specification.

There is very little work that needs to be done on this PR to finish it. We had discussed how the tests could be completed on last Tuesdays call. We were expecting you on the Thursday Special Topic call to answer any questions that you may have had and to move things forward, but you were not in attendance.

Please see the notes from the call here:

https://www.w3.org/2021/04/29-did-topic-minutes.html#t01

Again, please:

1) Confirm that you will finish these tests.
2) Finish the tests by next Tuesday (ideally, they'd be done today or tomorrow ... the sooner the better).

csuwildcat:
I'll have the PR updated by then

msporny:
I'm doing the final review of tests in the test suite... I just noticed that all of the JSON Consumption tests are missing (this PR doesn't cover those, I thought it did). 

@csuwildcat please make sure you impelement tests for everything under 6.2.1 [JSON] Production and 6.2.2 [JSON] Consumption in the following issue:

https://github.com/w3c/did-test-suite/issues/75

Every single one of those tests needs to be implemented and working before the next DID WG call.

packages/did-core-test-server/suites/did-production/did-json-production.js
```diff
@@ -1,16 +1,77 @@
...
+
+  it('6.2.1 JSON Production - The DID document and any DID document ' +
+    'data structures expressed by the data model MUST be serialized to the ' +
+    'JSON representation according to the JSON representation production ' +
+    'rules as defined in § 6.2 JSON.', async () => {
+      reserializedDidDocument = JSON.parse(JSON.stringify(didDocument));
+      expect(deepEqual(didDocument, reserializedDidDocument)).toBe(true);
+  });
+
+  it('6.2.1 JSON Production - When serializing a JSON-LD ' +
```
msporny, line 18:
```suggestion
  it('6.2.1 JSON Production - When serializing a JSON ' +
```

msporny (commented):
> @msporny @OR13 please let me know what you think about that latest PR addition, as I honestly still only have half a clue (HT to Orie, because previously I had 0) about whether this is what I was supposed to do.

The atomic tests are more or less correct, but the structure of the test suite is wrong. :)

You need every single one of these to be an `it(...)` test (NOTE: Please put the section title as a prefix... that is "6.2.1 JSON Production - " should go before each test... so "6.2.1 JSON Production - integer: A JSON Number without a decimal or fractional component."):

### 6.2.1 [JSON] Production

* [x] The DID document, DID document data structures, and representation-specific entries map MUST be serialized to the JSON representation according to the following production rules:
* [ ] map: A JSON Object, where each entry is serialized as a member of the JSON Object with the entry key as a JSON String member name and the entry value according to its type, as defined in this table.
* [ ] list: A JSON Array, where each element of the list is serialized, in order, as a value of the array according to its type, as defined in this table.
* [ ] set :A JSON Array, where each element of the set is added, in order, as a value of the array according to its type, as defined in this table.
* [ ] datetime: A JSON String serialized as an XML Datetime normalized to UTC 00:00:00 and without sub-second decimal precision.
* [ ] string: A JSON String.
* [ ] integer: A JSON Number without a decimal or fractional component.
* [ ] double: A JSON Number with a decimal and fractional component.
* [ ] boolean: A JSON Boolean.
* [ ] null: A JSON null literal.
* [ ] All entries of a DID document MUST be included in the root JSON Object.
* [ ] When serializing a DID document, a conforming producer MUST specify a media type of application/did+json to downstream applications such as described in § 7.1.2 DID Resolution Metadata.

### 6.2.2 [JSON] Consumption

* [ ] The DID document and DID document data structures JSON representation MUST be deserialized into the data model according to the following consumption rules:
* [ ] JSON Object: A map, where each member of the JSON Object is added as an entry to the map. Each entry key is set as the JSON Object member name. Each entry value is set by converting the JSON Object member value according to the JSON representation type as defined in this table. Since order is not specified by JSON Objects, no insertion order is guaranteed.
* [ ] JSON Array where the data model entry value is a list or unknown: A list, where each value of the JSON Array is added to the list in order, converted based on the JSON representation type of the array value, as defined in this table.
* [ ] JSON Array where the data model entry value is a set: A set, where each value of the JSON Array is added to the set in order, converted based on the JSON representation type of the array value, as defined in this table.
* [ ] JSON String where data model entry value is a datetime:  A datetime.
* [ ] JSON String, where the data model entry value type is string or unknown: A string.
* [ ] JSON Number without a decimal or fractional component: An integer.
* [ ] JSON Number with a decimal and fractional component, or when entry value is a double regardless of inclusion of fractional component: A double.
* [ ] JSON Boolean: A boolean.
* [ ] JSON null literal: A null value.
* [ ] If media type information is available to a conforming consumer and the media type value is application/did+json, then the data structure being consumed is a DID document, and the root element MUST be a JSON Object where all members of the object are entries of the DID document.
* [ ] A conforming consumer for a JSON representation that is consuming a DID document with a root element that is not a JSON Object MUST report an error.

Please see this for an example: 

https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/did-production/did-jsonld-production.js#L24-L40

csuwildcat:
@msporny this doesn't make sense to me. I have tests in there now for all known, top-level properties, and we're already testing to make sure the values parse correctly. If the JSON variant doesn't care about anything else, why would we be testing for unused props that the JSON variant doesn't even care about?

csuwildcat:
The part that's bananas to me is that:

1. We start off by verifying the inbound document is JSON, so obviously no non-JSON types are coming in
2. We then test for all known props, which are the only ones the JSON representation cares about
3. Then we need to test for what? Random props we aren't aware of and aren't even looking for or using in the JSON representation? Are you saying I need to add an example object with random, made-up props that don't get used in JSON, then test them for value matches?

csuwildcat:
Why don't we just test this whole thing against a JSON schema, so that it can pick up every prop and test for whatever we want? I don't really understand the value in setting some random, contrived prop on a DID Doc object, like `didDocument.foo = 123`, and then having code that tests that prop is a number. Is that what people are asking for here? What value does this add?

csuwildcat:
OK, switched the whole thing to a loop that checks only known props, which means there is literally no way you would ever be testing some of these types. I think this clarifies the question at the heart of this: It doesn't make sense to test things that won't be dealt with, because we have already validated that it is JSON, thus legal props for inclusion, but not needing to be value tested.

packages/did-core-test-server/suites/did-production/did-json-production.js
csuwildcat: Covered in PR updates

msporny:
@csuwildcat these tests are all still wrong - we can't use them as they stand. Please re-read https://github.com/w3c/did-test-suite/pull/71#pullrequestreview-649610781 -- we need an `it()` statement for every checkbox there.

To be clear -- we are testing the implementability of the normative statements in the specification. To do that, we have to have a test for every single normative statement. All of the normative statements that apply to the test you are writing are listed here: https://github.com/w3c/did-test-suite/pull/71#pullrequestreview-649610781 

The approach you're describing sounds more like you're expecting to write some sort of linting tool (JSON Schema). We can't do that and get out of CR. We need to test every machine-testable normative statement.

msporny:
To ensure that the JSON Production and Consumption sections are not dropped from the DID Core specification, thus kicking off yet another syntax scruffle in the WG, I have implemented all of them here: https://github.com/w3c/did-test-suite/pull/83

@csuwildcat -- you are welcome to continue with this PR if you disagree with the direction of the other PR. Letting you know that this is done now in case you wanted to disengage.

csuwildcat:
I just scanned your tests and it seems those are exclusively about testing
values of properties for type compliance, but don't we still want to retain
tests that recognize and act on the expected properties of a DID Document?
I would think it's useful to a developer to understand if the shape/props
conform to not only high-level types, but also the correct content for a
given property. Is this not desired?

msporny:
> I just scanned your tests and it seems those are exclusively about testing values of properties for type compliance, but don't we still want to retain tests that recognize and act on the expected properties of a DID Document? I would think it's useful to a developer to understand if the shape/props conform to not only high-level types, but also the correct content for a given property. Is this not desired?

Yes, that's desired, and is what the "Core Properties" suite of tests is about:

https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/did-core-properties/did-core-properties.js

Those were completed a while ago and are, thus, not necessary for the production/consumption tests.

csuwildcat:
@msporny yeah, those are the kinds of tests I thought I was writing, which is why I was confused about certain value type tests, because they would never happen with the registered props that the JSON variant knows about. I am sorry for my confusion, and I guess I will close this if those other kinds of tests are already done as well.

msporny:
PR #83 has been merged, which supersedes this PR. Closing.

Issue #77, opened 2021-04-27
Normative statements checklist: 8. Methods


rhiaro:
_NOTE: none of these are machine testable as they involve looking over DID method spec text, so I don't expect any to be checked off as part of the test suite, but including for completeness._

## 8.1 Method Syntax

* [ ] A DID method specification MUST define exactly one method-specific DID scheme that is identified by exactly one method name as specified by the method-name rule in § 3.1 DID Syntax.
* [ ] The DID method specification MUST specify how to generate the method-specific-id component of a DID.
* [ ] The DID method specification MUST define sensitivity and normalization of the value of the method-specific-id.
* [ ] The method-specific-id value MUST be unique within a DID method. The method-specific-id value itself might be globally unique.
* [ ] Any DID generated by a DID method MUST be globally unique.
* [ ] To reduce the chances of method-name conflicts, a DID method specification SHOULD be registered in the DID Specification Registries [DID-SPEC-REGISTRIES].
* [ ] The method-specific-id format MAY include colons. The use of colons MUST comply syntactically with the method-specific-id ABNF rule.
* [ ] A DID method specification MUST define how authorization is performed to execute all operations, including any necessary cryptographic processes.
* [ ] A DID method specification MUST specify how a DID controller creates a DID and its associated DID document.
* [ ] A DID method specification MUST specify how a DID resolver uses a DID to resolve a DID document, including how the DID resolver can verify the authenticity of the response.
* [ ] A DID method specification MUST specify what constitutes an update to a DID document and how a DID controller can update a DID document or state that updates are not possible.
* [ ] The DID method specification MUST specify how a DID controller can deactivate a DID or state that deactivation is not possible.
* [ ] A DID method specifications MUST follow all guidelines and normative language provided in RFC3552: Writing Security Considerations Sections for the DID operations defined in the DID method specification.
* [ ] The Security Considerations section MUST document the following forms of attack for the DID operations defined in the DID method specification: eavesdropping, replay, message insertion, deletion, modification, denial of service, storage or network amplification, and man-in-the-middle. Other known forms of attack SHOULD also be documented.
* [ ] The Security Considerations section MUST discuss residual risks, such as the risks from compromise in a related protocol, incorrect implementation, or cipher after threat mitigation was deployed.
* [ ] The Security Considerations section MUST provide integrity protection and update authentication for all operations required by Section § 8.2 Method Operations.
[...]

Issue #72, opened 2021-04-27
Normative statements checklist: Section 3. Identifiers


rhiaro:
## 3.1 DID Syntax

* [x] All DIDs MUST conform to the DID Syntax ABNF Rules.

## 3.2 DID URL Syntax

* [x] All DID URLs MUST conform to the DID URL Syntax ABNF Rules.

### 3.2.1 DID Parameters 

* [x] `service`: If present, the associated value MUST be an ASCII string.
* [x] `relativeRef`: If present, the associated value MUST be an ASCII string and MUST use percent-encoding for certain characters as specified in RFC3986 Section 2.1.
* [x] `versionId`: If present, the associated value MUST be an ASCII string.
* [x] `versionTime`: If present, the associated value MUST be an ASCII string which is a valid XML datetime value, as defined in section 3.3.7 of W3C XML Schema Definition Language (XSD) 1.1 Part 2: Datatypes [XMLSCHEMA11-2].
* [x] ...This datetime value MUST be normalized to UTC 00:00:00 and without sub-second decimal precision.
* [x] `hl`: If present, the associated value MUST be an ASCII string.
* [ ] For maximum interoperability, it is RECOMMENDED that DID parameters use the DID Specification Registries mechanism [DID-SPEC-REGISTRIES]

### 3.2.2 Relative DID URLs

* [x] When resolving a relative DID URL reference, the algorithm specified in RFC3986 Section 5: Reference Resolution MUST be used.

msporny:
I have performed a full review of the tests above. All tests with ticked checkboxes exist in the test suite. All tests that are not checked do not have active/associated tests in the test suite.

PR #80, opened 2021-04-30
Add remaining missing tests that came up during test suite audit.

msporny:
This PR adds tests to the test suite that were missing based on a review of issues #72 - #77.

### DIFF
--- a/packages/did-core-test-server/suites/did-consumption/default.js
+++ b/packages/did-core-test-server/suites/did-consumption/default.js
@@ -1,53 +1,10 @@
 module.exports = {
-  conformingConsumers: [{
-    "name": "non-conforming DID",
-    "input": {
-      "mediaType": "application/did+ld+json",
-      "representation": "{\"id\":\"not-a-did:true\",\"@context\":[\"https://www.w3.org/ns/did/v1\"]}"
-    },
-    "output": {
-      "didDocumentDataModel": {
-        "id": "not-a-did:true"
-      },
-      "representationSpecificEntries": {
-        "@context": ["https://www.w3.org/ns/did/v1"]
-      },
-      "errors": [
-        "invalidDid"
-      ]
-    }
-  },
-  {
-    "name": "non-conforming DID Document",
-    "input": {
-      "mediaType": "application/did+ld+json",
-      "representation": "{\"@context\":[\"https://www.w3.org/ns/did/v1\"]}"
-    },
-    "output": {
-      "didDocumentDataModel": {},
-      "representationSpecificEntries": {
-        "@context": ["https://www.w3.org/ns/did/v1"]
-      },
-      "errors": [
-        "missingId"
-      ]
-    }
-  },
-  {
-    "name": "did:example:123",
-    "input": {
-      "mediaType": "application/did+ld+json",
-      "representation": "{\"id\":\"did:example:123\",\"@context\":[\"https://www.w3.org/ns/did/v1\"]}",
-      "options": {}
-    },
-    "output": {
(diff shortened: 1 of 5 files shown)


packages/did-core-test-server/suites/did-identifier/did-syntax.js
```diff
@@ -8,19 +8,6 @@ const didSyntaxTests = (suiteConfig) => {
...
-        });
-        it('The DID method name MUST be an ASCII lowercase string.', () => {
-          const method = didExample.split(':')[1];
-          expect(method).toBeAsciiString();
-          expect(method.toLowerCase()).toBe(method);
-        });
-        it('The DID method name MUST NOT be empty.', () => {
-          const method = didExample.split(':')[1];
-          expect(method).not.toBe('');
-        });
```
msporny, line 23: The removal of these tests are correct -- they were replaced w/ new test language in the spec (and a test that covers all three of these cases).

packages/did-core-test-server/suites/did-consumption/default.js
```diff
@@ -1,53 +1,10 @@
 module.exports = {
-  conformingConsumers: [{
```
msporny, line 2: Removal of this and replacement with actual DID Methods is correct -- old consumer tests were using old consumer data (and not using implementations).

packages/did-core-test-server/suites/did-consumption/did-jsonld-consumption.js
```diff
@@ -0,0 +1,40 @@
+const deepEqual = require('deep-equal')
```
shigeya, line 1: Some of the tests using deepEqual, but there is a standard matcher checking for equality: `toBeEqual`.
Of course, the results are the same. Maybe we can refactor later.

FYI: [What is the difference between 'toBe' and 'toEqual' in Jest?](https://stackoverflow.com/questions/45195025/what-is-the-difference-between-tobe-and-toequal-in-jest)

packages/did-core-test-server/suites/did-consumption/did-jsonld-consumption.js
```diff
@@ -0,0 +1,40 @@
...
+    ...didDocumentDataModel.representationSpecificEntries
+  };
+  const {representation} = resolutionResult;
+
+  it('6.3.2 JSON-LD Consumption - The DID document and any DID document ' +
+    'data structures expressed by a JSON-LD representation MUST be ' +
+    'deserialized into the data model according to the JSON ' +
+    'representation consumption rules as defined in § 6.2 JSON.', async () => {
+      const consumedDidDocument = JSON.parse(representation);
+      expect(deepEqual(didDocument, consumedDidDocument)).toBe(true);
```
shigeya, line 16: For example, this can be written as:
> expect(didDocument).toBeEqual(consumedDidDocument);

shigeya (approved):

OR13 (approved):

packages/did-core-test-server/suites/did-consumption/did-jsonld-consumption.js
msporny: Yes, I will refactor this in a future PR. We should be using the .toEqual() test in Jest. I didn't know it existed at the time I pulled in deep-equal.

packages/did-core-test-server/suites/did-consumption/did-jsonld-consumption.js
msporny: Yes, will refactor later.

PR #83, opened 2021-05-02
Add JSON Production and Consumption tests.

msporny:
This PR adds JSON Production and Consumption tests as well as extending `did:example` in `application/did+json` format to include all possible values in the data model (datetime, integer, double, boolean, and null).

### DIFF
--- a/packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
+++ b/packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
@@ -79,7 +79,7 @@ const generateJsonConsumptionTests = (
   it('6.2.2 JSON Consumption - A conforming consumer for a JSON ' +
     'representation that is consuming a DID document with a root ' +
     'element that is not a JSON Object MUST report an error.', () => {
-    expect(typeof didDocument === 'object');
+    expect(typeof didDocument === 'object').toBeTruthy();
   });
 
   it('6.2.2 JSON Production - If media type information is available to ' +
--- a/packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
+++ b/packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
@@ -18,14 +18,14 @@ const generateJsonConsumptionTests = (
           'Array is added to the list in order, converted based on the JSON ' +
           'representation type of the array value, as defined in this ' +
           'table.', () => {
-          expect(reserializationSuccess(value)).toBe(true);
+          expect(consume(value)).toBe(true);
         });
         it('6.2.2 JSON Consumption - JSON Array where the data model entry ' +
           'value is a set: A set, where each value of the JSON Array is ' +
           'added to the set in order, converted based on the JSON ' +
           'representation type of the array value, as defined in this ' +
           'table.', () => {
-          expect(reserializationSuccess(value)).toBe(true);
+          expect(consume(value)).toBe(true);
         });
       } else if(value !== null && typeof value === 'object') {
         it('6.2.2 JSON Consumption - JSON Object: A map, where each member ' +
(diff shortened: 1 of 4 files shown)


peacekeeper (approved):

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
```diff
@@ -0,0 +1,133 @@
...
+    'application/did+json, then the data structure being consumed is a ' +
+    'DID document, and the root element MUST be a JSON Object where all ' +
+    'members of the object are entries of the DID document.', async () => {
+      const {contentType} = resolutionResult.didResolutionMetadata;
+      expect(contentType).toBe('application/did+json');
+  });
+}
+
+const reserializationSuccess = (value) => {
+  const reserialization = JSON.parse(JSON.stringify(value));
```
shigeya, line 97: Is it safer to wrap for an exception, even we know its structure is well-formed?  e.g.,  https://github.com/w3c/did-test-suite/blob/385760b56a7627294bfff04831e63887a0101b87/packages/did-core-test-server/suites/did-core-properties/did-metadata-structure.js#L42-L51

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
```diff
@@ -0,0 +1,133 @@
...
+          'set by converting the JSON Object member value according to the ' +
+          'JSON representation type as defined in this table. Since order ' +
+          'is not specified by JSON Objects, no insertion order is ' +
+          'guaranteed.', () => {
+          expect(reserializationSuccess(value)).toBe(true);
+        });
+      } else if(typeof value === 'string' && isXmlDatetime(value)) {
+        it('6.2.2 JSON Consumption - JSON String where data model entry ' +
+          'value is a datetime: A datetime.', () => {
+          expect(reserializationSuccess(value)).toBe(true);
```
shigeya, line 43: This test does not test whether the value is valid DateTime string or not.

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
```diff
@@ -0,0 +1,133 @@
...
+        it('6.2.2 JSON Consumption - JSON Boolean: A boolean.', () => {
+          expect(reserializationSuccess(value)).toBe(true);
+        });
+      } else if(value === null) {
+        it('6.2.2 JSON Consumption - JSON null literal: A null value.', () => {
+          expect(reserializationSuccess(value)).toBe(true);
+        });
+      } else {
+        it('UNKNOWN JSON TYPE '+ value, () => {
+          expect(true).toBe(true);
```
shigeya, line 74: This should be an error.

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
```diff
@@ -0,0 +1,133 @@
...
+        });
+      }
+    });
+  });
+
+  it('6.2.2 JSON Consumption - A conforming consumer for a JSON ' +
+    'representation that is consuming a DID document with a root ' +
+    'element that is not a JSON Object MUST report an error.', () => {
+    expect(typeof didDocument === 'object');
```
shigeya, line 83:
```suggestion
    expect(typeof didDocument === 'object').toBeTruthy();
```

shigeya (approved):
While there is room for improvement, these tests seem reasonable.

There is one error better be addressed.

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
msporny: I looked into this. Ideally, we'd use a Jest test for this, but I couldn't figure out exactly what to write. I think it's fine for now. We can always refactor later. If an exception is thrown, Jest will say where since we're in a test when it happens.

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
msporny: It does. It has to be a valid XML Datetime for the test to fire. The consumption test then checks for round-tripping. I do agree that it's not a straightforward test, but the check is being performed (even though it's not obvious it is).

packages/did-core-test-server/suites/did-consumption/did-json-consumption.js
msporny: Fixed in 550fc1bd9b494843995c31c021047fcc54fd29a1.

shigeya (approved):

shigeya (approved):

PR #85, opened 2021-05-03
Tiny improvement to XML datetime matcher.

peacekeeper:


packages/jest-did-matcher/src/matchers/toBeDidCoreDatetime/predicate.js
```diff
@@ -1,5 +1,5 @@
 export default expected => {
-    const regex = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z/
+    const regex = /[0-9]{4}-[0-1][0-9]-[0-3][0-9]T[0-2][0-9]:[0-6][0-9]:[0-6][0-9]Z/
```
shigeya, line 2: We don't observe 23:60:60 ?

```suggestion
    const regex = /[0-9]{4}-[0-1][0-9]-[0-3][0-9]T[0-2][0-9]:[0-5][0-9]:[0-5][0-9]Z/
```

shigeya (changes requested):

packages/jest-did-matcher/src/matchers/toBeDidCoreDatetime/predicate.js
peacekeeper: Of course, thanks :)

shigeya (approved):

OR13 (approved):

shigeya:
I now realized that the regexp only test is naive. We can improve further by using the date parser. Possibly faster?

packages/jest-did-matcher/src/matchers/toBeDidCoreDatetime/predicate.js
```diff
@@ -1,5 +1,4 @@
 export default expected => {
-    const regex = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z/
+    const regex = /[0-9]{4}-[0-1][0-9]-[0-3][0-9]T[0-2][0-9]:[0-5][0-9]:[0-5][0-9]Z/
```
msporny, line 2: I suggest we use the official definition for the XMLDateTime regex:

https://www.w3.org/TR/xmlschema11-2/#nt-dateTimeRep

This is not something we should be attempting to create on our own.

peacekeeper: I agree, but I don't know how to create an EBNF parser in JavaScript, or if some existing library can be used for parsing XML dateTime. So someone else would have to do that...

msporny: > I agree, but I don't know how to create an EBNF parser in JavaScript

That specification has a regex that you can copy/paste into JavaScript... look further down the page.

msporny:
> I now realized that the regexp only test is naive. We can improve further by using the date parser. Possibly faster?

I believe the Javascript Date() parser allows non-XMLDateTimes as well... it's not as strict as the Regex, IIRC.

shigeya:
> > I now realized that the regexp only test is naive. We can improve further by using the date parser. Possibly faster?
> 
> I believe the Javascript Date() parser allows non-XMLDateTimes as well... it's not as strict as the Regex, IIRC.

Right. I meant we need both regexp match and JavaScript Date() parser to check the ranges of each of the fields. 
Since range check code is error-prone, it better rely on JavaScript Date() parser.

msporny:
> Right. I meant we need both regexp match and JavaScript Date() parser to check the ranges of each of the fields.
> Since range check code is error-prone, it better rely on JavaScript Date() parser.

Ah! I see... yes, you're correct @shigeya -- that would be a better check.

peacekeeper:
Note that in DID Core we have the additional constraint that the DateTime has to be _"normalized to UTC 00:00:00 and without sub-second decimal precision"_.

I don't think I know how to write the code to properly test all of the above myself. Maybe I close this PR and create an issue instead?

shigeya:
> Note that in DID Core we have the additional constraint that the DateTime has to be _"normalized to UTC 00:00:00 and without sub-second decimal precision"_.
> 
> I don't think I know how to write the code to properly test all of the above myself. Maybe I close this PR and create an issue instead?

What Manu pointed out was just parsing the string by Date parser allows submilliseconds.

Since the current regexp does not allow to include submilliseconds, and also enforcing UTC (timezone letter `Z`), additional checking also with Data parser (numerical range check) seems better.

I can write a PR if you want. either fine.

peacekeeper:
> I can write a PR (possibly a follow-up to this PR) if you want. either fine.

That sounds good, I think you would be more efficient in doing this than me.

msporny:
@peacekeeper, what are we doing with this PR? I think you're suggesting to close it, above?

PR #93, opened 2021-05-08
chore: Fix test for 3ID.

oed:
This PR fixes the did-resolution test for `did:3` and moves the test vectors using DID URLs over to the dereferencer test.

### DIFF
--- a/packages/did-core-test-server/suites/did-resolution/default.js
+++ b/packages/did-core-test-server/suites/did-resolution/default.js
@@ -1,8 +1,8 @@
 module.exports = {
   "name": "7.1 DID Resolution",
   "resolvers": [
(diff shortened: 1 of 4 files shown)


peacekeeper (approved):
@oed don't get me wrong, just curious.. Did you actually update your implementation, or did you simply "fix the tests manually"? :)

oed:
@peacekeeper The [did-resolver](https://github.com/decentralized-identity/did-resolver) doesn't yet distinguish between "resolving" and "dereferencing", so this is mostly just reshuffling of the test vectors here. I imagine that the did-resolver package will get update once the spec solidifies. 
To be clear: All DID documents in the test vectors are generated from the actual implementation.

PR #96, opened 2021-05-10
Add Universal Resolver implementations

BernhardFuchs:
- All drivers which return a `200` Status code
- `did+ld+json` representation
- only `defaultOutcomes`


BernhardFuchs:
@OR13 @msporny This is done from our side.  
It includes good cases for all Universal Resolver drivers which return a `200`.  

The files are generated in a semi-automatic process with these [scripts](https://github.com/decentralized-identity/universal-resolver/tree/driver-status-reporting/ci/run-did-test-suite) in the `ci`. The idea here is to have a ci-step to run the results of the Universal Resolver against a hosted version of the `did-test-suite` (probably a cron job). I'm struggling a bit with the input parameters to the `/generate-report` endpoint. See https://github.com/w3c/did-test-suite/issues/104

OR13:
@BernhardFuchs what is remaining to move this from draft to mergable?

BernhardFuchs:
We have just added two more (ebsi and sol) examples.  
Otherwise this is ready to merge.

msporny:
@BernhardFuchs Please fix merge conflicts...

BernhardFuchs:
@msporny Conflicts are resolved

shigeya:
Hi, the current test run is failing after the merge of this commit. I raised issue #111 please kindly check the issue.

Note that I found this file is missing: `implementations/universal-resolver-did-factom.json`
Please kindly provide the file. Maybe, as a PR?

picardcapt1212:
Hello, i have a question. 
We created "did:lit" method and registered our resolver in universal resolver.
While processing the did test suite we realized our did method should be changed, 
which means universal resolver example did document should also change.

But i checked our "did:lit" resolver is already merged as "implementations/universal-resolver-did-lit.json".
I think if we update "did:lit" method, "didDocumentStream" field in "implementations/universal-resolver-did-lit.json" will slightly change. 

How should i update this? Thank you.

BernhardFuchs:
@picardcapt1212 We are currently creating `implementations`-files for all working drivers on the Universal Resolver but that doesn't mean we are the authorithy over them. So feel free to create a PR with the updated `implementations/universal-resolver-did-lit.json` or create your own file in the `implementations` folder.  

The only issue I see here, is that we have a semi-automated process to create those files. In case you have updated the `implementations/universal-resolver-did-lit.json` file but the driver for the Universal Resolver still produces the outdated result, we can't guarantee that the `universal-resolver-did-lit.json` doesn't get overwritten with deprecated data. That's why we prefixed all the Universal Resolver related files with `universal-resolver-did-`.

picardcapt1212:
> @picardcapt1212 We are currently creating `implementations`-files for all working drivers on the Universal Resolver but that doesn't mean we are the authorithy over them. So feel free to create a PR with the updated `implementations/universal-resolver-did-lit.json` or create your own file in the `implementations` folder.
> 
> The only issue I see here, is that we have a semi-automated process to create those files. In case you have updated the `implementations/universal-resolver-did-lit.json` file but the driver for the Universal Resolver still produces the outdated result, we can't guarantee that the `universal-resolver-did-lit.json` doesn't get overwritten with deprecated data. That's why we prefixed all the Universal Resolver related files with `universal-resolver-did-`.


Okay, then i guess if we update `lit-resolver` we have and also update `implementations/universal-resolver-did-lit.json` there will be no problem.. 

Thank you for reply!

PR #101, opened 2021-05-11
Fix regex for toBeValidDid/isValidDid to match with ABNF rules in latest spec

etaleo:
The ABNF in https://www.w3.org/TR/did-core/#did-syntax allows for DIDs like `did:example:123abc:123abc`, while the current didRegex does not, which should be fixed by this PR

You can also verify the regex works as intended via https://regex101.com/


packages/jest-did-matcher/src/matchers/toBeValidDid/predicate.js
```diff
@@ -1,4 +1,4 @@
 export default expected => {
-    const didRegex = /^did:(?<method>[a-z0-9]+):(?<idchar>([a-zA-Z0-9.\-_]|%[0-9a-fA-F]{2})+$)/;
+    const didRegex = /^did:(?<method>[a-z0-9]+)(:(?<idchar>([a-zA-Z0-9.\-_]|%[0-9a-fA-F]{2})+))+$/;
```
kdenhartog, line 2: Looks good, but this regex has no length upperlimit allowing it to be susceptible to a ReDoS attack. I'm concerned this regex will get copied from this test into an implementation. Should we be setting a max limit on the size of a DID?

etaleo: I have also realized that the ABNF allows for DIDs like did:example::::::abc:::123, while my regex does not, so it might need further tweaking.

A quick and dirty approach would be something like this (so it now matches 1:1 with the ABNF):
`^did:(?<method>[a-z0-9]+):((?<idchar>([a-zA-Z0-9.\-_]|%[0-9a-fA-F]{2})*):)*(?<idchar2>([a-zA-Z0-9.\-_]|%[0-9a-fA-F]{2})+)$`

But here I just copied & pasted the idchar-group, which is not ideal. Certainly one could come up with a cleaner regex, maybe by constructing it over multiple lines instead of cramping it into one line (or using some group-reusing inside the same regex, which would sacrifice readbility).

Your concerns regarding a ReDos attack seem valid, though I'm unsure of the implications on this (would one need to add this limit into the did-core spec itself? Or would it be okay to differ from the did-core spec there?)

kdenhartog:
Tagging @peacekeeper and @talltree who are the ABNF/URI experts who may want to weigh in here.

msporny:
@etaleo -- the regex was updated in a separate PR. Please check it and see if your PR is needed any longer.

etaleo:
> @etaleo -- the regex was updated in a separate PR. Please check it and see if your PR is needed any longer.

The regex included in [this PR](https://github.com/w3c/did-test-suite/pull/97/commits/5a0df8c62066c9178f0c957f288fdd3799334e22#diff-6e8967c37688e9138b80b129ccf5aff92f9d1be1e217a5968d8827a4889748e6) does allow for DIDs like `did:example:abc:::`, while the ABNF requires that it does not end with a `:` (but [with idchar instead](https://www.w3.org/TR/did-core/#did-syntax)).

I think when we add an additional `[^:]` before the end the regex should fully align with the ABNF:

```^did:(?<method_name>[a-z0-9]+):(?<method_specific_id>([a-zA-Z0-9.\-_]|%[0-9a-fA-F]{2}|:)+[^:]$)```

msporny:
> _edit:_ This does not allow for DIDs like `did:example:abc%00`, so further tweaking is needed

Noted -- would you mind putting in a PR to fix all these issues @etaleo -- we're not being bitten by the restrictive regex yet, but I imagine we will in time.

shigeya:
Thanks for the fix. If possible, please include test cases. If not, I will add tests later.

msporny (approved):

msporny:
Merge conflicts, please fix, @etaleo.

OR13:
Can this be merged.?

shigeya (approved):

msporny:
Squash merge it is.

TallTed:
@shigeya @msporny @etaleo --

There were several "DIDs like ..." examples that were flagged above as not allowed for by various formulations of the regex, which I think are supposed to be allowed for now -- e.g., `did:example:abc%00`, `did:example::::::abc:::123` -- and at least one *negative* test -- `did:example:abc:::` -- but none of these samples are in the validation DIDs I see near the regex.

I think these should be added there.

shigeya:
@TallTed Thanks for the head-up. I created a PR to add test

@msporny wrote:
>> edit: This does not allow for DIDs like did:example:abc%00, so further tweaking is needed
> Noted -- would you mind putting in a PR to fix all these issues @etaleo? We're not being bitten by the restrictive regex yet, but I imagine we will in time.

The regexp still lacks the support of the above pattern.

I tried to implement using lookahead regexp, ended up with a simpler solution with two regexps.
PR created. Please review.

Issue #78, opened 2021-04-29
Report generation is brittle and needs some development work


brentzundel:
We have need for additional developer resources to help finish the automatic generation of test implementation report.

Much work has been done: 
https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/package.json#L11https://github.com/w3c/did-test-suite/tree/main/packages/did-core-test-server#usagehttps://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/generate-latest-respec-data.js
https://github.com/w3c/did-test-suite/blob/main/index.html#L686

Please comment here if you may be able to help, and reach out to @OR13 for more detail.

aljones15:
Have you tried a custom reporter? I've written two in Mocha and the last one has actually turned out to be pretty useful and painless to use. [Jest Custom Reporter docs](https://jestjs.io/docs/configuration) [Mocha w3c interop reporter](https://github.com/digitalbazaar/mocha-w3c-interop-reporter). The later produces the nightly reports for the vaccine credentials interoperability report.

OR13:
@aljones15 

https://transmute-industries.github.io/did-core/
https://github.com/transmute-industries/did-core/blob/main/packages/did-core-tests/jest.config.js#L6

But these don't solve the problem of an NxN breakdown where you can count feature support....

Luckily this exists:

https://github.com/digitalbazaar/vaccination-certificate-test-suite
http://vaxcert-interop-reports.s3-website.us-east-2.amazonaws.com/#COVID-19

^ maybe you can add some support for something like this? Built on the existing JSON which the test suite produces?

Here is an example: https://github.com/w3c/did-test-suite/blob/main/test-vectors/did-spec.latest.json

shigeya:
@OR13 I still don't understand how I can generate a report. How can I configure `./suites/did-spec/default.json` ?

OR13:
This line updates the latest json for the report in the html:

https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/generate-latest-respec-data.js#L10

This line loads the json from the html which is describes the latests run:

https://github.com/w3c/did-test-suite/blob/main/index.html#L869

Looks like I forgot that this PR never got merged....

https://github.com/w3c/did-test-suite/pull/20/files#diff-9893d200a81ade372ac94c51de4364d040a15fa75250a8b8be6a742d6d4e0697R2

@msporny @peacekeeper ^ perhaps this would have helped with report rendering.

[...]

PR #105, opened 2021-05-16
Initial set of report generator update

shigeya:
Initial work for report generation. Related to Issue #78.

I'm sending this PR to provide the latest status of this work and wish to have feedbacks on the report structure since I'm changing that a lot.

- Suites configuration is now in `suites/suite-config.json`. Currently, only provides the list of test suites. The configuration parameters in this file override configurations in each of the suites. Currently, each of the `default.js` files contains defaults for each of the suites. I think these should be moved to `suite-config.json`
- All of the tests tweaked to have the same root ancestors suite-name and implementation name, to make report matrix creation easier. These were included in part of the describe statement, but it's not easily machine-parsable and some of them are inconsistent.
- All of the long raw outputs are moved to Appendix.
- Added/updated Terminology section.
- Replaced with new "Conformance Report" section with three subsections: Matrix, by Test Suites, by Methods.

***
<a href="https://pr-preview.s3.amazonaws.com/shigeya/did-test-suite/pull/105.html" title="Last updated on May 18, 2021, 3:45 PM UTC (3e557c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-test-suite/105/271a4f9...shigeya:3e557c4.html" title="Last updated on May 18, 2021, 3:45 PM UTC (3e557c4)">Diff</a>


shigeya:
@OR13 @msporny Please provide feedback. I'm now working on conformance section generation.

shigeya:
Remaining tasks:

- [ ] Conformance Report Section
    - [ ] Matrix
    - [ ] by Suites
    - [ ] by Methods

Also, 
- [ ] Decides to keep On-demand section or not

shigeya:
First reworked version for "Conformance by Test Suites" section.

I lack of CSS kung-fu. Maybe, we need adjustments on fonts to be used in the statements and tables.

msporny:
@shigeya, this is wonderful, thank you!

A couple of thoughts:

* We shouldn't check in index.html or did-spec.latest.json -- we shouldn't check in auto-generated files, instead we should invest in running the test suite on a regular basis, like we did for the vaccination test suite: https://github.com/digitalbazaar/vaccination-certificate-test-suite/runs/2429248801?check_suite_focus=true.

shigeya:
> @shigeya, this is wonderful, thank you!

thanks.

> * The initial report looks great, it's formatted so that it's easy to understand the details. It was what I was hoping someone would produce, so this looks wonderful!

Good to hear that.

> * It is currently difficult to see which features don't have enough support. For example, the JSON `null` test only has one implementation that has implemented it. It would be good to see these in a section that lists the "Features Lacking at least Two Implementations".

Looking through the test report again, it seems there are some cases not handled correctly. Will check them.

> * I expect we shouldn't count did:example in the list of implementations? However, if we don't do this, there are a number of JSON-only features that would be dropped.

We need to remove examples from the report.
Maybe, either keep them as part of tests (then skip from the output) or keep them as an example configured to skip tests, etc.

> * It's difficult to understand why `alsoKnownAs` has so many implementations supporting it... specifically, Digital Bazaar's implementations don't currently support the feature but it still shows our implementations as supporting `alsoKnownAs`. That leads me to believe that their might be a bug in the tallying code.

Will check the code.

> * We shouldn't check index.html or did-spec.latest.json into the github repo -- we shouldn't check in auto-generated files, instead we should invest in running the test suite on a regular basis and publishing to a website, like we did for the vaccination test suite: https://github.com/digitalbazaar/vaccination-certificate-test-suite/runs/2429248801?check_suite_focus=true... this file is generated as a part of the build process: https://w3id.org/vaccination/interop-reports

I agree. Actually, that is my preference. Since I'm relying on Orie's code I kept the flow of the report generation. The current report page contains too much automation in the report. Since the number of tests entry is increasing, the report is getting larger and slower, that's what we don't want to see.

If it's okay, I'll move most of the code from the index.html into the report generator. Also, remove the report JSON in the report, and maybe removing the On Deman Test section.

shigeya:
Consolidated parameters into a single column, also showing parameters that show how each of the tests is different.
Looks like `did-resolution` and `did-url-dereferencing` sections are still strange, all of the other tables look good. 

@msporny 
> It's difficult to understand why alsoKnownAs has so many implementations supporting it... specifically, Digital Bazaar's implementations don't currently support the feature but it still shows our implementations as supporting alsoKnownAs. That leads me to believe that their might be a bug in the tallying code.

it was not tallying problem but the test code on 5.1.3.
The test code block `it` succeeds even there is no `alsoKownAs` property.
Please refer to the commit 4cf9626a8590503124f9c51d9d7f623d96ddbb3a

But now, the empty table causes the table for `alsoKnownAs` test to disapper. Need to find a way to fix this.

shigeya:
I will work on moving most of the report generation code in index.html into the report generator code, then good to merge. We can improve later. I don't want to away from up-to-date `main` branch. I don't have time for this in the next 24hrs, so it will be ready after Tuesday's call.

I leave resolution/derefencing as is for now.

shigeya:
I cherry-picked PR #110 and PR #111. Will remove them when this PR exits DRAFT.

shigeya:
> I leave resolution/dereferencing as is for now.

Fixed.

I'll move the code in HTML to the report generator script next.

OR13:
this looks awesome! great work on this.

shigeya:
It was easy. Now it is generating a report by `generate-latest-respec-data.js`.

Besides matrix generation, it's done. please review and merge.

For the testing purpose, need PR #110 to test to pass ... to generate a report.

packages/did-core-test-server/services/runSuite.js
```diff
@@ -8,7 +8,7 @@ module.exports = async (suiteName, config) => {
       // ci: true,
       // silent: true,
       roots: [path.resolve(__dirname, `../suites/${suiteName}`)],
-      globals: JSON.stringify({ suiteConfig: config }),
+      globals: JSON.stringify({ systemSuiteConfig: config }),
```
OR13, line 11: for folks wondering, this is the magic of our jest tests... we are injecting a global json object here... which is the same object that an http client can submit to an http web server to generate the report json.

OR13 (approved):
Thanks for this PR, it looks excellent.

packages/did-core-test-server/services/runSuite.js
shigeya: Indeed. It was a shock to understand that `globals` is a jest's option. The name `globals` is obvious from the feature, but the name is strange for an optional parameter.

OR13:
This PR has been open for 6 days and has multiple approvals, merging.

PR #117, opened 2021-05-21
chore: add orb did

fqutishat:
list of did samples:

- did:orb:interim:EiDHdXhNm7LuCqxo4JvAwKYKiFmpf85YFswAovxTxI_y4Q (interim did, local cas, V1 CID, P-256 JsonWebKey2020, empty service)

- did:orb:bafkreiazah4qrybzyapmrmk2dhldz24vfmavethcrgcoq7qhic63zz55ru:EiAag4cmgxAE2isL5HG3mxjS7WRq4l-xyyTgULCAcEHQQQ (canonical did, local cas, V1 CID, P-256 JsonWebKey2020, empty service)

- did:orb:bafkreihp4inweep4py7gw4j7hej5mqlbwa7br4u7mtrfxr5khfwpu3qu3m:EiB2tmdM_oWwjXj6AmVLm0RFa_8XKZHipOpNGpEODIVN8Q (canonical did, local cas, V1 CID, ed25519 Ed25519VerificationKey2018, DIDComm v1 service)

- did:orb:bafkreibcsubh3ifub7gletz27hcdyhwvrhlh5mwfth2m5fbasqua6yalay:EiA2ZtZqXjKZt-yf19ersmaCYm-gJEnlixrfk0Mi61ETTg (canonical did, local cas, V1 CID, P-256 JsonWebKey2020, DIDComm v2 service)

- did:orb:webcas:testnet.orb.local:bafkreihdnftiso5b7bzmhhi65nzsutbcuv6mtrmuquzoqlrk7joyer45uq:EiARiEOCLK3GnRVHA_yF92tX3aoSJAVqW1bh7Enre1iDXw (webcas did, local cas, V1 CID, P-256 JsonWebKey2020, DIDComm v2 service)

- did:orb:interim:EiCYgffdSsqLTXT6PRYLPr6vvgn9PVecJ5nFUGh9hXgOxQ (interim did, ipfs, V1 CID, P-256 JsonWebKey2020, DIDComm v2 service)

- did:orb:ipfs:bafkreiacr3ga6zilvzatpcixq5mz4uvgld7yedutgcssvnmql44o6rc7yy:EiB2k0ytmo-qi_M7jGocxvj4P9D6VQJGl6gRy4f6-UUpTw (ipfs did, ipfs, V1 CID, P-256 JsonWebKey2020, DIDComm v2 service)

[...]

Issue #112, opened 2021-05-19
didMetadataStructureTest_r is not defined


fqutishat:
```
    ReferenceError: didMetadataStructureTest_r is not defined

      24 |             Object.keys(didDocumentMetadata).forEach((property_name) => {
      25 |                 if (typeof didDocumentMetadata[property_name] == "object") {
    > 26 |                     didMetadataStructureTest_r(didDocumentMetadata[property_name]);
         |                     ^
      27 |                 }
      28 |             });
      29 |         });

      at forEach (suites/did-core-properties/did-metadata-structure.js:26:21)
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (suites/did-core-properties/did-metadata-structure.js:24:46)
```

https://github.com/w3c/did-test-suite/blob/333b1220483d31d723004a036e1c74e0243d1aac/packages/did-core-test-server/suites/did-core-properties/did-metadata-structure.js#L26

fqutishat:
I think there is typo in method name should be didMetadataStructureTests

shigeya:
@fqutishat Thank you very much for spotting this. This code path is not used unless you run with a property with an object as a value.

 If you need to run the test, please rename

`didMetadataStructureTest_r`

to 

`didMetadataStructureTest_Map_Recursive`

Then, it should work.

I will create PR for the repository.

shigeya:
@fqutishat The current head of the main branch merged the fix. please try. 

@msporny Please mark this issue as pending close.

msporny:
> @msporny Please mark this issue as pending close.

We only need to do that on DID Core... we can close issues on the test suite as they're confirmed. All we need is @fqutishat to say you fixed his issue and we can close.

fqutishat:
@msporny @shigeya I have this error now 
```
 ● 7.3 Metadata Structure - did:orb - DID Orb Test Suite - SecureKey › did:orb:interim:EiC-dLIbtnkHnfc6_Wkc8-X8Ag8FhvpC-jIMrT29_S_ttw › application/did+ld+json › 7.3 Metadata Structure - The values within any complex data structures such as maps and lists MUST be one of these data types as well.

    Tests cannot be nested. Test "7.3 Metadata Structure - The structure used to communicate this metadata MUST be a map of properties." cannot run because it is nested within "7.3 Metadata Structure - The values within any complex data structures such as maps and lists MUST be one of these data types as well.".

[...]

PR #122, opened 2021-05-22
Addresses metadata test failure on Issue #112

shigeya:
This is a follow-up PR that addresses issues mentioned in Issue #112 to support the test of multi-level map structure in DID Metadata tests using the new `toBeDidDocumentMap` matcher.

It passes most of the tests in the `did-core-properties` test suite for `did:orb` mentioned in PR #117.


shigeya:
A few fixes related to this PR added also.

Issue #142, opened 2021-05-31
Type error on endpoint


BernhardFuchs:
When running the did-test-suite server from main, the `/test-suite-manager/generate-report` throws an error about wrong type if provided, with a now correct, object.

PR #143, opened 2021-05-31
Use type object instead of array for endpoint

BernhardFuchs:
fixes https://github.com/w3c/did-test-suite/issues/142

### DIFF
--- a/packages/did-core-test-server/routes/index.js
+++ b/packages/did-core-test-server/routes/index.js
@@ -23,7 +23,7 @@ const registerRoutes = (server) => {
         summary: "Generate Report",
         description: "Run tests on supplied suites",
         body: {
-          type: "array",
+          type: "object",
           example: suitesInput,
           additionalProperties: true,
         },


msporny:
@shigeya any idea if this PR is going to cause problems w/ other inputs?

@BernhardFuchs I'm a bit lost as to what input was causing an error before?

shigeya:
@msporny `route/index.js` is outside of the non-server test run and report generation, so it will not cause any effect to test and report generation.

I'm also curious about the reason for the change.

BernhardFuchs:
@msporny

Before the [changes](https://github.com/w3c/did-test-suite/commit/8a2fa15be0a4c2462b2699274b5b384ba053c551#diff-10f21c106611de6331b858d4873fdbcddc129078a0f460c9980d428302a3006c) to [getReportResults.js](https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/services/getReportResults.js) the input was constructed as an `Array`:

```
[
  {
    "name": "did-resolution",
    "resolvers": [
      // implementations go here in the same format as the static files in the implementations folder
    ]
  }
]
```

Now it has to be an `Object` of `suites` which contains the test-suites:
```
{
  "suites":
    [
      {
        "suite_name": "did-resolution",
        "resolvers": [
          // implementations go here in the same format as the static files in the implementations folder
        ]
      }
    ]
}
```
Atm the endpoint rejects the `Object` because it still expects an `Array`.

shigeya:
@BernhardFuchs thanks for the clarificaiton. I have totally forgotten about that change.

shigeya (approved):

PR #145, opened 2021-05-31
Add transmute implementations

OR13:
- [x] Fix CI 
- [x] Add Transmute implementations


OR13:
@msporny as long as "only passing" is green, you can safely merge PRs, since the assumption is that the tests are now stable enough to be relied on for CI purposes.

Each suite has some fixtures which cause CI to fail, and you can review the logs in "include breaking" to see the culprits.

ideally, we would not have a CI that fails, and include breaking could be removed.

It would also be nice, if CI generated the report, but that would require additional work, (see https://github.com/w3c/did-test-suite/issues/146)

msporny:
I have concerns related to this PR: https://github.com/w3c/did-test-suite/pull/145/commits/fb70316ab2d50b267fbb136c6c63b67d97fd7d7f ... it looks like people's implementations were commented out... is that still going on in this PR? If so, that's a problem -- this test suite isn't intended to block people from submitting implementations that are not conformant... since we need to be able to do that to ensure that the specification is reflecting reality. I couldn't tell from a cursory look if we're stopping implementers from being a part of the implementations if they don't 100% pass everything.

Any chance we can split out Transmute's implementations from the CI work? I don't understand why this PR bunches both of them together.

.gitignore
```diff
@@ -7,7 +7,7 @@
 TAGS
 config.json
 node_modules
-package-lock.json
+
```
msporny, line 10: We don't need to check in package-lock.json files -- things are changing rapidly still.

OR13:
> Any chance we can split out Transmute's implementations from the CI work? I don't understand why this PR bunches both of them together.

I needed to fix the tests, so I could tell out implementation was not at fault for them failing... I expect others would need to do the same.

I can split the CI changes out, but all they really do is add an env flag in front of the registered fixtures.

Looks like I might have missed on though.

.gitignore
OR13: package dependencies should not be rapidly changing, and if they are, package-lock should be committed to avoid ambiguity... https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

OR13:
As you can see, CI is still failing when all fixtures are included, yet the additions in this PR are not failing.

msporny (approved):
For the record, I hate package-lock.json files when used outside of production software packaging cycles. They've been the source of many a developer hour wasted. :)

That said, LGTM, merging.

msporny:
@OR13 do you want to clean up your commit history, or do you want me to squash-merge? I'd prefer the former, but can do the latter.

OR13:
@msporny I prefer squash and merge.

PR #147, opened 2021-06-01
add ION resolution file

csuwildcat:


msporny (changes requested):
You need to add this resolver to the resolution test default.js file:

https://github.com/w3c/did-test-suite/blob/main/packages/did-core-test-server/suites/did-resolution/default.js

msporny (approved):

PR #148, opened 2021-06-01
Add bid DID Method to the did-test-suite

teleinfo-bif:
Added the bid DID method for the following suites:

did-consumption
did-core-properties
did-identifier
did-production


msporny (changes requested):
Looks like you forgot to include the implementation files?

PR #161, opened 2021-06-06
Add deno-did-pm dereference example

OR13:
This PR makes use of a couple rarely used did core feature:

- service
- relativeRef
- dereference (with non did core content type, in this case typescript)

The example is pulled from this did package manager for the deno programing language:

https://github.com/OR13/deno-did-pm

since deno uses URLs for modules, any deno module can be represented as a DID URL on a web origin that supports did dereferencing.

The same is likely true of GoLang as well.


OR13:
@teleinfo-bif heads up, you broke the CI when this was merged

https://github.com/w3c/did-test-suite/commit/5b18e24c5568a176586fa6ac0ea310a1728a4cb4

You need to escape the JSON in that fixture properly.

I I have moved it to the broken fixtures list, in order to get my PR to pass CI.

msporny (approved):

PR #177, opened 2021-06-30
Deactivate tests that are not actually based on normative statements.

peacekeeper:
This comments out some tests that are interesting but are not actually based on normative statements in the specification. Those can potentially be added back in the future once there are more implementations.

This PR also improves the description strings of some tests, so it's clear which property a test is testing.

### DIFF
--- a/packages/did-core-test-server/suites/did-resolution/did-resolution.js
+++ b/packages/did-core-test-server/suites/did-resolution/did-resolution.js
@@ -183,33 +183,37 @@ const didResolutionTests = (execution, expectedOutcome, imp) => {
 const testDidDocumentMetadata = (didDocumentMetadata, didDocument, expectedOutcome) => {
   describe('created', () => {
     if (didDocumentMetadata.hasOwnProperty('created')) {
-      it('7.1.3 DID Document Metadata - The value of the property MUST be a string formatted as an XML Datetime normalized to UTC 00:00:00 and without sub-second decimal precision.', async () => {
+      it('7.1.3 DID Document Metadata - created - The value of the property MUST be a string formatted as an XML Datetime normalized to UTC 00:00:00 and without sub-second decimal precision.', async () => {
         expect(didDocumentMetadata['created']).toBeDidCoreDatetime();
       });
     }
   });
   describe('updated', () => {
     if (didDocumentMetadata.hasOwnProperty('updated')) {
-      it('7.1.3 DID Document Metadata - The value of the property MUST follow the same formatting rules as the created property.', async () => {
+      it('7.1.3 DID Document Metadata - updated - The value of the property MUST follow the same formatting rules as the created property.', async () => {
         expect(didDocumentMetadata['updated']).toBeDidCoreDatetime();
       });
+      /*
+       * While it is interesting to test this, it's not actually a normative statement in the specification, therefore this doesn't have to be tested.
+       *
       if (didDocumentMetadata.hasOwnProperty('created')) {
-        it('7.1.3 DID Document Metadata - updated is later or equal than created.', async () => {
+        it('7.1.3 DID Document Metadata - updated - updated is later or equal than created.', async () => {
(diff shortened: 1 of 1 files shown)


OR13 (approved):

shigeya:
@peacekeeper It looks like you changed the descriptions but not deactivating the tests. Is this your intention?

peacekeeper:
@shigeya this PR does two things:

1. Slightly change the descriptions of many tests
2. Deactivate 3 tests, by commenting them out `/* .... */`

shigeya (approved):

shigeya:
> 2. Deactivate 4 tests, by commenting them out `/* .... */`

Found the `/*` ... `*/` pair. I see and thanks for the trouble.

msporny (changes requested):
Whoa, red flag -- totally disagree with this PR.

These are *normative* MUST statements in the specification... it doesn't matter if the properties are optional, their optionality isn't what's being tested... what's being tested is that if they exist they MUST be X.

I'm marking this as `do not merge`, the WG will have to discuss this if we're going to merge the PR because it will be going back on multiple previous resolutions and decisions we made to test all normative MUST statements in the specification (which underwent multiple reviews wrt. which statements we're testing).

kdenhartog (approved):
Good catch @peacekeeper looks like these are actually normative statements in the spec

peacekeeper:
@msporny this PR deactivates the following four tests:

- 7.1.3 DID Document Metadata - updated - updated is later or equal than created.
- 7.1.3 DID Document Metadata - nextUpdate - nextUpdate is later than created.
- 7.1.3 DID Document Metadata - nextUpdate - nextUpdate is later than updated.
- 7.1.3 DID Document Metadata - nextVersionId - nextVersionId is different from versionId.

None of them are normative statements in the spec. Maybe those tests should have never been written in the first place.

Other than that, the PR also slightly changes descriptions of many other tests, but it does NOT deactivate those other tests.

Do you still see a problem with this PR?

msporny:
> Do you still see a problem with this PR?

Ah, I see. Ok, so the tests are testing normative language that no longer exists in the specification (or never existed in the first place). I'm going to put some time aside to do a line-by-line check to make sure we're good here... it's always tricky changing the test suite this late in the game. If we do the wrong thing, we could end up in a 3rd Candidate Recommendation phase (because of an objection that we didn't test all normative statements, which is unlikely as long as we really are removing non-normative statements).

I'll do another review and cross-check and then merge if there are no issues. @shigeya @OR13 @kdenhartog -- did each of you cross-check against the spec that the tests being commented out don't exist as normative statements in the specification?

OR13:
@msporny I have not confirmed that the spec has not changed its normative requirements... in part because programmatically detecting that had been disabled last I checked.

In our tests, we covered 100% of the the sentences... but then when did core makes changes, this breaks this..... https://github.com/transmute-industries/did-core/tree/main/packages/did-core-tests#generate-json-test-cases

... So we stopped checking... I am in general OK not testing DID Resolution Response Objects, other than the DID Document serialization, I find them to be generally poorly defined beyond the abstract data model... I would suggest they be covered in DID Resolution tests for the DID Resolution spec.

peacekeeper:
> I'll do another review and cross-check and then merge if there are no issues.

Thanks. I'm pretty sure those have never been normative statements in the spec, and they were only added to the test suite because they "felt reasonable". But yes it can't hurt if someone other than me also checks this, just to make sure.

msporny:
> None of them are normative statements in the spec. Maybe those tests should have never been written in the first place.

I have now confirmed that all of the tests being deactivated in this PR are not actual normative requirements in the spec.

>  they were only added to the test suite because they "felt reasonable".

Yep, that should never have been done. The test suite is supposed to test all normative MUST statements and that's it. Looks like these tests got past quality control. :)

msporny (approved):
I have confirmed that this PR updates tests in the spec in an appropriate way and removal of tests only remove things that are not normative statements in the specification today.

kdenhartog:
Just seeing this now, but yeah I had manually checked the spec before approving. Thanks for checking as well @msporny

PR #186, opened 2021-07-15
Add implementation of did:ebsi method.

peacekeeper:
This adds an implementation report for `did:ebsi`. It was generated using libraries and APIs from the [EBSI](https://ec.europa.eu/cefdigital/wiki/display/CEFDIGITAL/EBSI) Early Adopter program. 

Generating this report did NOT use any parts of the DIF Universal Resolver (even though that also supports `did:ebsi`).

### DIFF
--- a/packages/did-core-test-server/suites/did-consumption/default.js
+++ b/packages/did-core-test-server/suites/did-consumption/default.js
@@ -36,6 +36,7 @@ module.exports = {
     require('../implementations/did-onion-spruce.json'),
     require('../implementations/did-pkh-spruce.json'),
     require('../implementations/did-webkey-spruce.json'),
+    require('../implementations/did-ebsi.json'),
     ...brokenFixtures
   ]
 };
--- a/packages/did-core-test-server/suites/did-core-properties/default.js
+++ b/packages/did-core-test-server/suites/did-core-properties/default.js
@@ -35,6 +35,7 @@ module.exports = {
     require('../implementations/did-onion-spruce.json'),
     require('../implementations/did-pkh-spruce.json'),
     require('../implementations/did-webkey-spruce.json'),
+    require('../implementations/did-ebsi.json'),
     ...brokenFixtures
 
   ],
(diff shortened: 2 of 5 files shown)


clehner (approved):

msporny:
@peacekeeper Is there public code (an implementation library / software application) that demonstrates the implementation of this feature? Given that we only have two implementers for this feature, it's helpful to point the Director to code if it comes up.

msporny:
I would still like to ensure that the EBSI early adopter program code exists -- would be easy if it were public on github or something.

Assuming it does:

1. Evernym created a DID Document that used `alsoKnownAs` and resolved it through the Universal Resolver (#187).
2. Danube Tech created a DID Document that used `alsoKnownAs` using libraries and APIs from the EBSI Early Adopter program (#186).

That is two independent implementations that have demonstrated support for `alsoKnownAs`.

I'll merge both to the test suite tomorrow, and remove the at risk issue marker for `alsoKnownAs`. :)

... but please do follow up and elaborate more on the code about EBSI, @peacekeeper .

kdenhartog:
@msporny I'd venture to guess it can be found somewhere in here. I haven't dug through it to find it though.

 https://ec.europa.eu/cefdigital/code/repos?visibility=public

peacekeeper:
@msporny your summary of the two implementations is accurate.

I don't think the EBSI code is public yet outside of the Early Adopter program, so I don't feel comfortable sharing it right now; but if there is any doubt about its existence, I could show it via screenshare, etc.

Also, you can verify that the DID document in this implementation report came from EBSI by executing this curl command:

```
curl -X 'GET' \
  'https://api.preprod.ebsi.eu/did-registry/v2/identifiers/did%3Aebsi%3A4RPpNqqGv6NtrMMuPStGUJQ4zgCmUn5ZNqcxmkmdpWbd' \
  -H 'accept: application/did+ld+json'
```

kdenhartog (approved):

msporny:
> Also, you can verify that the DID document in this implementation report came from EBSI by executing this curl command:

The curl command is "good enough" -- that we can hit two live systems, run by different organizations, and get a DID Document that has `alsoKnownAs` in it is a defensible position to the Director. Thanks, I'll merge both and remove the at risk marker from `alsoKnownAs`.

PR #196, opened 2021-07-26
Remove todo test for untestable statement

### LINKED ISSUE #194: Symbols are confusing
In the implementation report, all normative statements are shown to be passing with green check marks except for [specification statement 3.2.2](https://w3c.github.io/did-test-suite/#T8) which has yellow triangles, but no mention is made in the document what this means.

It is clear that green check marks are good, and red x's are bad, but what does the yellow triangle mean? Is specification statement 3.2.2 implemented or not?

@shigeya @OR13 @msporny

brentzundel:
Fixes #194 
Signed-off-by: Brent Zundel <brent.zundel@gmail.com>

### DIFF
--- a/packages/did-core-test-server/suites/did-url-dereferencing/did-url-dereferencing.js
+++ b/packages/did-core-test-server/suites/did-url-dereferencing/did-url-dereferencing.js
@@ -14,7 +14,6 @@ const didUrlDereferencingTests = (execution, expectedOutcome, imp) => {
           }
         });
   
-        it.todo('3.2.2 Relative DID URLs - When resolving a relative DID URL reference, the algorithm specified in RFC3986 Section 5: Reference Resolution MUST be used.');
         it('7.2 DID URL Dereferencing - A conformant DID URL as a single string.', async() => {
           if (! dereferencingMetadata.hasOwnProperty('error') || dereferencingMetadata['error'] !== 'invalidDidUrl') {
             expect(didUrl).toBeValidDidUrl();


OR13 (approved):

shigeya (approved):

kdenhartog:
If we're removing this test do we also need to remove the statement from DID Core?

https://w3c.github.io/did-core/#relative-did-urls

```
When resolving a relative DID URL reference, the algorithm specified in RFC3986 Section 5: Reference Resolution MUST be used. The base URI value is the DID that is associated with the DID subject, see § 5.1.1 DID Subject. The scheme is did. The authority is a combination of <method-name>:<method-specific-id>, and the path, query, and fragment values are those defined in § Path, § Query, and § Fragment, respectively. 
```

OR13:
@kdenhartog no.

OR13:
@msporny please review, I would like to merge

msporny (approved):

kdenhartog:
> @kdenhartog no.

Can you comment on why not? If it's because of the practicality of impact on the phases of the spec that's fine with me. I guess my concern is that it's a normatively defined statement, so we should either determine it's untestable and modify/remove the statement in the spec (in a follow on WG) or we figure out how to test it. Removing it from the test suite and not the spec doesn't make much sense to me.

msporny:
> Removing it from the test suite and not the spec doesn't make much sense to me.

Oh, crap. I thought folks had already made sure this statement wasn't in the spec... that'll teach me to not triple-check. Good catch, @kdenhartog!

@kdenhartog is absolutely correct... this is a normative, testable statement in the specification. It never got implemented as a test... and my scan for all implemented features missed that this wasn't implemented because doing a `.todo()` reported that it was implemented (it was in the test suite).

So, let this be a lesson to all of us -- we can't remove .todo()s w/o confirming that the statement isn't in the spec anymore, and we probably need to do an audit of all `.todo()`s that were removed to make sure we didn't miss anything.

I'm backing this commit out of the test suite. It is a `.todo()` and we can't remove it. It's a gap in the test suite, it is testable, so we need to implement it for DID Dereferencer conformance.

PR #205, opened 2021-11-05
Update DID EBSI method

yhuard:
The DID EBSI method has been recently redefined, causing some breaking changes. This PR provides a valid example of a DID document resolved by querying https://api.preprod.ebsi.eu/did-registry/v2/identifiers/did:ebsi:znHeZWvhAK2FK2Dk1jXNe7m.

⚠️ The project is still in beta, so we can't guarantee there won't be breaking changes in the future.

Note: the lockfile was automatically updated when I first ran `npm i`. I have npm@8.1.0 installed locally.

```
❯ npm i
npm WARN old lockfile 
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8629 @@
 {
   "name": "root",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "root",
+      "version": "0.0.1",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "devDependencies": {
+        "lerna": "^3.22.1"
+      },
+      "engines": {
+        "node": ">=14"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.12.11",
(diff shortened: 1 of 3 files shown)


msporny (changes requested):
Please drop the package-lock.json file from the commit and either 1) raise a separate PR to update it, or 2) drop it entirely and we'll do it.

peacekeeper:
Note that `did:ebsi` isn't registered yet in https://w3c.github.io/did-spec-registries/#did-methods.

yhuard:
@msporny I've dropped the package-lock.json file from the PR. I'll let you handle it.

@peacekeeper it's probably better to keep it under the radar for the time being, since EBSI is still in development

msporny (approved):

msporny:
@yhuard -- this PR broke the "Only Passing" build, please fix it -- https://github.com/w3c/did-test-suite/runs/4355760571?check_suite_focus=true

msporny:
> @msporny #206 should fix it ok_hand

It did, thank you! :)

PR #227, opened 2025-09-06
Implement DID v1.1 tests.

msporny:
This PR is an attempt to upgrade the DID test suite to support v1.1 (which uses a separate context entry). The purpose here is to effectively duplicate all existing implementations, upgrading them to support v1.0 and v1.1 DID Documents. This PR proves that approach to be viable. 

We might also choose to do JSON-LD canonicalization on the v1.1 documents to ensure they canonicalize to the same values as v1.0.

### DIFF
--- a/packages/did-core-test-server/suites/did-production/did-jsonld-production.js
+++ b/packages/did-core-test-server/suites/did-production/did-jsonld-production.js
@@ -19,7 +19,7 @@ const generateJsonldProductionTests = (
       expect(didDocument).toHaveProperty('@context');
   });
 
-  it('6.3.1 JSON-LD Production - The serialized value of @context ' +
+  it('6.3.1 JSON-LD Production (DID v1.0) - The serialized value of @context ' +
     'MUST be the JSON String https://www.w3.org/ns/did/v1, or a JSON ' +
     'Array where the first item is the JSON String ' +
     'https://www.w3.org/ns/did/v1 and the subsequent items are serialized ' +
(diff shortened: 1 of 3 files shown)


packages/did-core-test-server/suites/did-production/did-jsonld-production.js
```diff
@@ -20,16 +20,18 @@ const generateJsonldProductionTests = (
...
+    'and the subsequent items are serialized ' +
     'according to the JSON representation production rules.', async () => {
       const context = didDocument['@context'];
 
       if(typeof context === 'string') {
-        expect(context).toBe('https://www.w3.org/ns/did/v1');
+        expect(context).toMatch(/^https\:\/\/www.w3.org\/ns\/did\/v1|https\:\/\/www.w3.org\/ns\/did\/v1\.1$/);
       } else if(Array.isArray(context)) {
-        expect(context[0]).toBe('https://www.w3.org/ns/did/v1');
+        expect(context[0]).toMatch(/^https\:\/\/www.w3.org\/ns\/did\/v1|https\:\/\/www.w3.org\/ns\/did\/v1\.1$/);
```
dlongley, line 34: If `toBe()` was the wrong assertion because it was doing something like checking a type, then maybe there is a `toEqual()` instead of `toMatch()` to simplify.

packages/did-core-test-server/suites/utils.js
```diff
@@ -105,6 +105,52 @@ const isValidVerificationMethod = (vm) => {
...
+
+    if(key === 'representation') {
+      let newRepresentation = JSON.parse(obj['representation']);
+      if(newRepresentation['@context'] !== undefined) {
+        if(newRepresentation['@context'][0] ===
+          'https://www.w3.org/ns/did/v1') {
+            newRepresentation['@context'][0] = 'https://www.w3.org/ns/did/v1.1';
+        }
+        obj['representation'] = JSON.stringify(newRepresentation);
+    }
```
dlongley, line 129:
```suggestion
      }
```

dlongley (approved):

packages/did-core-test-server/suites/did-production/did-jsonld-production.js
msporny: The problem is that jest doesn't support matching multiple strings, it's either a single value, you pull in a separate library that adds a  `.oneOf()` match, or you use a regex. I went with a regex because pulling in the extra library would require an update to package-lock.json, which would trigger a whole bunch of upgrades necessary across the entire test infrastructure.

The current infrastructure is pretty terrible -- it's over-engineered and in need of a significant re-write. Once we have did-resolution working, we should probably use that instead of the current test suite. That would require significant effort to re-write, so we're having to make suboptimal choices in order to reduce this cycle of testing.

packages/did-core-test-server/suites/utils.js
```diff
@@ -105,6 +105,52 @@ const isValidVerificationMethod = (vm) => {
...
+  }
+}
+
+const addDidv11Implementations = (implementations) => {
+  const allImplementations = [];
+
+  implementations.forEach((suiteConfig) => {
+    v10SuiteConfig = suiteConfig;
+    v11SuiteConfig = JSON.parse(JSON.stringify(suiteConfig));
```
davidlehn, line 143: Is it ok to use the standard clone api here?
```suggestion
    v11SuiteConfig = structuredClone(suiteConfig);
```

msporny: Yes. I was unsure if it was available in node.js 20, which we have to use for this test suite. Looks like it's been in node.js since node.js 17.

